PR #8, opened 2022-01-31
Impl Erase Spell

Noobulus:
also add a colorizer because i wanted one


gamma-delta:
nepotism

Noobulus:
i deserve this for fucking up git so hard

PR #13, opened 2022-02-05
Fix ForEach not adding the element to the stack at the start of each loop

Hudeler:
Thoth's Gambit was broken since the rewrite as it didn't add the elements it iterates over to the stack at the start of each loop. This is a simple fixed since it was just a case of re-adding a line from the previous implementation.


gamma-delta:
Thanks!

PR #14, opened 2022-02-05
Fix error message thrown by OpForEach

Hudeler:
Added a check for the number of elements on the stack at the start of OpForEach. For now, the number of args is hardcoded, which should probably be changed to a more elegant solution.
Previously the error messages were wrong, as getChecked passes the index that was out of bounds to the NOT_ENOUGH_ARGS CastException, instead of the number of args which are required. It's unclear how to fix it, as getChecked doesn't actually know how many args are required, so this should probably be cleaned up as well.
Therefore, this is more a temporary fix, but it works for now.


gamma-delta:
I suppose we'll pull this until someone else yells at us

PR #15, opened 2022-02-05
proper translation components, now available

Jeuvke:
I did not test this at all so it might be a bit scuffed. Translation components of inputs/outputs for spells will come in a later PR.


gamma-delta:
You've been busy!

PR #27, opened 2022-03-13
Ridcully's Projection

12foo:
Made this as a test spell to get familiar with how spells work in Hex, feel free to include or throw away! It's a spell that fires a spectral arrow. (For perhaps obvious reasons, from a crossbow.)


12foo:
Actually the trick is that it's a spectral arrow, so you can use it to find mobs through walls. But yep, perfectly understandable!

PR #68, opened 2022-05-08
Add cons lists for list destructuring & more efficient list ops

Alwinfy:


yrsegal (changes requested):
Update to latest main branch (i.e. the change to casting using SpellDatums instead of HexPatterns)

PR #71, opened 2022-05-14
ru_ru.json lang.

LedinecMing:


gamma-delta:
Ooh fancy! ... We're *about* to do a giant refactor though, so I would hold on a bit and rebase once we merge multiloader into main.

gamma-delta:
Awesome!

PR #85, opened 2022-05-22
Add Modular Math Operation

shBLOCK:


yrsegal:
mm. i'm not a fan of this pattern? Not the operation, just the pattern. And it's called modulus, not modular

PR #88, opened 2022-05-23
Rewrite the evaluator to trampoline side effects

Alwinfy:
TODO:
- [x] We can make Hermes/Thoth *not* stop at first mishap, and in fact keep chugging along with garbage on the stack! (Do we want this? Perhaps not.)
- [x] What's the pattern for Halt, actually?


Common/src/main/java/at/petrak/hexcasting/api/block/circle/BlockEntityAbstractImpetus.java
```diff
@@ -278,14 +278,15 @@ private void castSpell() {
                 new SpellCircleContext(this.getBlockPos(), bounds, this.activatorAlwaysInRange()));
             var harness = new CastingHarness(ctx);
 
+            var castSpell = false;
```
yrsegal: ???

yrsegal (approved):

Common/src/main/java/at/petrak/hexcasting/api/block/circle/BlockEntityAbstractImpetus.java
Alwinfy: epic merge dead code fail

Alwinfy:
Superseded by #89

PR #149, opened 2022-06-24
Hermes docupdate

KirinDave:
This updates the Hermes entry to reflect the new type signature an mention Charon's Gambit.

Infinidoge notes that my solution to current build problem is probably wrong, and that you might want to "move the hexcasting:jeweler_hammer entry into Common?" Please feel free to reject this PR if you want me to back out that change.


KirinDave:
With infinidoge submitting the build fix PRs to you, I should factor this more cleanly. Please disregard.

PR #183, opened 2022-07-24
[Fabric] Fix CardinalCoordinates-Related Crash, Edify Sapling, Fix Flight Duration, Improve Docs

### LINKED ISSUE #175: Fabric: Crashes on EdifySapling due to AkashicTreeGrower not registering during Initializer
[Crashlog examples](https://gist.github.com/gattsuru/cd74ead1b8f56750fbdeea058efcb196#file-gistfile1-txt)

Essentially, the [singleton here](https://github.com/gamma-delta/HexMod/blob/88d596cfd1c203683b9cb1992694a684805bd982/Common/src/main/java/at/petrak/hexcasting/common/misc/AkashicTreeGrower.java#L25) only gets called during a cast of EdifySapling, which is too late for the underlying registration at [this line](https://github.com/gamma-delta/HexMod/blob/88d596cfd1c203683b9cb1992694a684805bd982/Common/src/main/java/at/petrak/hexcasting/common/misc/AkashicTreeGrower.java#L36).

Adding the single line : 
``AkashicTreeGrower.INSTANCE``
at [FabricHexInitializer](https://github.com/gamma-delta/HexMod/blob/54724ac95e60b48d143ac79aec86c91ed224ea71/Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexInitializer.kt#L100) ln100 seems to fix it without any side effects for Fabric or Forge, though I can understand if that's a violation of the code style.

EDIT: I've got a branch with this modification and a bump to the Cardinal Components accepted versions [here](https://github.com/gattsuru/HexMod/tree/fix173+175), if it would be easier to look at that way, though I can understand not wanting to deal with this as a PR.

### LINKED ISSUE #176: [BUG] Flight spell works indefinitely
I haven't tested this much but on my friends fabric server when casting the flight spell I can cast it with values of 0 consuming 0 media and still be able to fly without any limits.

gattsuru:
- Fixes #175 : Bumps accepted range for Cardinal Components to ``>=4.1.4``, sets the Fabric build config to 4.2.0.  I'd rather 4.1.4-4.2, but I haven't been able to find compatible syntax.  You may want to just turn this into ``~4.2.0`` directly: the 4.14 version works in Single Player if nothing requires 4.2, but does cause sporadic crashes in multi-player remote.
- Instantiates EdifySapling's builder during initialization in Fabric, to prevent a related crash.
- Fixes #176 : Adds a tick handler for Fabric to handle Flight duration/distance.
- Improves documentation for Fabric Config files, since they seem to (sometimes?) crash on get with invalid values rather than something more helpful.
- Improves Impulse documentation re: increasing costs.


Common/src/main/resources/assets/hexcasting/lang/en_us.json
```diff
@@ -904,7 +904,7 @@
...
   "hexcasting.page.basic_spell.explode.fire.2": "Costs one $(l:items/amethyst)$(item)Amethyst Dust/$, plus about 3 extra $(l:items/amethyst)$(item)Amethyst Dust/$s per point of explosion power. Otherwise, the same as $(l:patterns/spells/basic#hexcasting:explode)$(action)Explosion/$, except with fire.",
-  "hexcasting.page.basic_spell.add_motion": "Remove an entity and direction from the stack, then give a shove to the given entity in the given direction. The strength of the impulse is determined by the length of the vector.$(br)Costs units of $(l:items/amethyst)$(item)Amethyst Dust/$ equal to the square of the length of the vector.",
+  "hexcasting.page.basic_spell.add_motion": "Remove an entity and direction from the stack, then give a shove to the given entity in the given direction. The strength of the impulse is determined by the length of the vector.$(br)Costs units of $(l:items/amethyst)$(item)Amethyst Dust/$ equal to the square of the length of the vector, plus one for every Impulse after the first in a spell.",
```
yrsegal: "besides" rather than "after" - this implies the cost compounds

Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexInitializer.kt
```diff
@@ -97,6 +100,8 @@ object FabricHexInitializer : ModInitializer {
 
         HexLootFunctions.registerSerializers(bind(Registry.LOOT_FUNCTION_TYPE))
 
+        AkashicTreeGrower.INSTANCE
```
gamma-delta: This is to just reify the class re: Java lazy initialization of statics? (Perhaps add a comment or something to that effect cause just seeing that sitting by itself looks like a typo without the context)

gamma-delta (changes requested):
Thanks a bundle! Would you mind commenting the bare `AkashicTreeGrower.INSTANCE` pls?

Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexInitializer.kt
yrsegal: Personally, I think using static initializers that need to be loaded at a given time should be turned into invoked functions

gattsuru: That's fair.  This approach had the smallest necessary change, but it definitely isn't obvious what it's doing.

Yeah, this is just to handle Java's lazy-initialization of AkashicTreeGrower, since its ``static{}`` call eventually uses ``FeatureUtils.register()``, which (in Fabric/Quilt and maybe vanilla?) has an explicit ``IllegalStateException("Registry is already frozen (trying to add key "")`` blocking that case.  Since there's no references to AkashicTreeGrower other than OpEdifySapling, it doesn't get called until someone casts that spell, and then the server side gives up.

Forge does allows those registrations available later, so the static init works fine in Forge cases.  But it's Not Good to load the same registry key twice; depending on modloader it will either crash or give a bunch of annoying [warn] messages.  The simple fix is to add the same registration invoke call into Forge's initializer, and that's what I've done for this commit.  

If that verboseness isn't ideal, you could keep the ``GROWERS.add`` calls in ``OpEdifySapling``'s ``static{}`` and then have an empty (or meaningless) ``RegisterTreeGrowers()`` that solely exists to instantiate the static variables.  I've had bad luck in the past with compiler optimizations pulling that sorta thing out, and it's not much clearer, though.

gamma-delta: I'd be OK with it as it is with a comment! That's what they're there for after all ...

gattsuru: yrsegal put together some pretty good options for ``init()`` functions, so the current commit uses them.  I can switch back and comment if you prefer that code style; it's really about what's going to be clearest for you all.

yrsegal:
What's the status on this?

gamma-delta (approved):

PR #184, opened 2022-07-24
[Fabric] Creates a Variable-Argument Set Gravity, Changes to Support That

gattsuru:
This one is more aggressive, and I'm not sure how much of it you will want to take.  In order:
- Separates the SpellOrder logic between consuming iotas from the stack, and the actual execution.  This allows individual spells to have more complex stack interactions without having to reimplement too much, or add complexity for normal spell args.
- Makes OpChangeGravity into:
image
- _I don't expect this to be the final implementation you'd want, if you even want these sort of changes at all_: it's intentionally harder than most normal spells to use for things like a Zone-based approaches using Thoth, and the prices are pretty much just made up wholesale.  A variant that allows casters to target themselves for a timed gravity effect, but also has the two-arg permanent version which mishaps if they use it on anyone else, is just as valid if you'd prefer that implementation.
- Adds the documentation shown above.
- Adds the various support structures necessary to allow that timer tickdown functionality.  Because the full set of all entities is much larger than that of all players, rather than the approach used in OpFlight, this has its own companion object with all entities with active timers.  I'm _not sure how good this is at consistency from server-to-client_; there's no obvious drop-outs in our play, but we haven't been hitting it hard yet and I don't really have a good understanding of how Cardinal Components handles a lot of edge cases.
- Kills players and discards entities that have been thrown too far (>5k blocks) into the sky.  This is probably overkill, but even with the timer it's pretty trivial to shoot someone into orbit, and this way there's a fix other than having an op kill you.

Some particular parts that I'm not hugely happy with:
- This adds a Client-side END_WORLD_TICK event that's basically just there to handle a case where a remote client player is gravity-changed, dies, and then the server think there's upside-up while the client doesn't get that update.  This _probably_ needs to be fixed better in GravityChanger/GravityAPI, but this approach is the best I could come up with on short notice without adding a lot of unnecessary network traffic, or adding a dependency on something like Architectury's ClientRespawnEvent.  Does not handle the connect-to-multiple-server or disconnected-when-timer-runs-out cases.
- Not entirely confident that the current Mishaps give enough insight, especially for Trinket use.  Intentional for Kirin's server, but not necessarily ideal for general use.
- The Forge implementation is basically just ??? right now.  GravityChanger doesn't have a Forge version, so I don't think that matters?
- ((GravityChanger itself probably needs some updating; along with the Client-Server confusion for a death while gravity-changed, there's a number of mixins we had to disable to stop it and Carpet from fighting over.  But even if that's updated, are likely to want to keep the fix implemented here til 1.19.))

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/misc/GravitySetting.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/misc/GravitySetting.java
@@ -1,0 +1,9 @@
+package at.petrak.hexcasting.api.misc;
+
+import net.minecraft.core.Direction;
+
+public record GravitySetting(Direction gravityDirection, boolean permanent, int timeLeft) {
+    public static GravitySetting deny() {
+        return new GravitySetting(Direction.DOWN, true, 0);
+    }
+}
--- a/Common/src/main/java/at/petrak/hexcasting/api/spell/SpellOperator.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/spell/SpellOperator.kt
@@ -17,11 +17,16 @@ interface SpellOperator : Operator {
         ctx: CastingContext
     ): Triple<RenderedSpell, Int, List<ParticleSpray>>?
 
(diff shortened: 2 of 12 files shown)


gamma-delta:
this is 4 months old so I'm gonna close it sorry

PR #305, opened 2022-11-21
Add missing properties to a few entries in the docs

object-Object:
Adds the following missing properties:

* Greater Sentinel: `anchor`
* Charon: `input`, `output`
* Consideration: `input`, `output`
* Introspection: `input`, `output`
* Retrospection: `input`, `output`

I generated the docs on my machine with these changes and everything seems to be working. I didn't build the mod because I don't have a dev environment set up, but I don't *think* these changes would break anything?

(I found these with [HexBug's type scraper](https://github.com/object-Object/HexBug/blob/main/scrape_book_types.py))

### DIFF
--- a/Common/src/main/resources/data/hexcasting/patchouli_books/thehexbook/en_us/entries/patterns/great_spells/greater_sentinel.json
+++ b/Common/src/main/resources/data/hexcasting/patchouli_books/thehexbook/en_us/entries/patterns/great_spells/greater_sentinel.json
@@ -9,6 +9,7 @@
     {
       "type": "hexcasting:pattern",
       "op_id": "hexcasting:sentinel/create/great",
+      "anchor": "hexcasting:sentinel/create/great",
       "text": "hexcasting.page.greater_sentinel.1",
       "input": "vector",
       "output": ""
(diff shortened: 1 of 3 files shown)


object-Object:
Also: Consideration, Introspection, Retrospection, and Novice's Gambit are all missing `op_id`, but I wasn't sure if adding that would break things, since the first three aren't in the registry and the last is actually Bookkeeper's Gambit.

gamma-delta:
Intro/retro/consideration shouldn't have inputs and outputs, and you're right those four shouldn't have op-ids (for exactly that reason)

If you fix those up I'll merge it

object-Object:
Re intro/retro/consideration: several other patterns also don't have inputs or outputs, but they still have the input and output fields, they're just empty strings. For example: [Summon Rain](https://github.com/gamma-delta/HexMod/blob/main/Common/src/main/resources/data/hexcasting/patchouli_books/thehexbook/en_us/entries/patterns/great_spells/weather_manip.json#L22)

gamma-delta:
Yeah, but still, intro/retro are *special*-er than summon rain... they're not really actions, they're more like syntax if anything. And it seems strange to annotate bits of syntax as if they were functions.

object-Object:
Makes sense, I guess. Updated

object-Object:
planning to fix this as part of the docgen pr

Issue #414, opened 2023-02-16
[WORKSPACE; 19.2 fabric] World Crash on lighting issue


Shibva:
"Bandaid" applied in commit linked below

## Information:

Version used: Workspace Development
Workspace Instance: 1.19.2
Launcher Instance/Version: Farbic
Workspace Itteration used: as of Commit 588a554efa436138ba067f9f608b1a10e2ca7c63

Issue: Upon world Load after a minute or so the workspace crashes wtih an error regarding a lighting issue

Additional Info: issue ceased when Phosphorus was disabled from the workspace

## Crash Report [GameLog if former is not available]: 
[crash-2023-02-15_17.18.43-client.txt](https://github.com/gamma-delta/HexMod/files/10760107/crash-2023-02-15_17.18.43-client.txt)
Report POI: ```Description: Unexpected error

java.lang.NoSuchMethodError: 'void net.minecraft.world.level.lighting.LayerLightEngine.method_15483(long)'
	at net.minecraft.world.level.lighting.LayerLightEngine.cancelUpdatesForChunk(LayerLightEngine.java:698)
	at net.minecraft.world.level.lighting.LayerLightSectionStorage.handler$beg000$preRemoveSection(LayerLightSectionStorage.java:816)
	at net.minecraft.world.level.lighting.LayerLightSectionStorage.clearQueuedSectionBlocks(LayerLightSectionStorage.java)
	at net.minecraft.world.level.lighting.LayerLightSectionStorage.addQueuedLightmaps(LayerLightSectionStorage.java:1423)
	at net.minecraft.world.level.lighting.LayerLightSectionStorage.markNewInconsistencies(LayerLightSectionStorage.java:852)
	at net.minecraft.world.level.lighting.LayerLightEngine.runUpdates(LayerLightEngine.java:186)
	at net.minecraft.world.level.lighting.LevelLightEngine.runUpdates(LevelLightEngine.java:63)
	at net.minecraft.client.renderer.LevelRenderer.renderLevel(LevelRenderer.java:1198)
	at net.minecraft.client.renderer.GameRenderer.renderLevel(GameRenderer.java:1039)
	at net.minecraft.client.renderer.GameRenderer.render(GameRenderer.java:816)
	at net.minecraft.client.Minecraft.runTick(Minecraft.java:1177)
	at net.minecraft.client.Minecraft.run(Minecraft.java:768)
	at net.minecraft.client.main.Main.run(Main.java:244)
[...]

PR #415, opened 2023-02-16
[Bandaid for #414] Disabled Phosphorus

Shibva:
World after a minute or so of joining will crash due to a lighting mishap [note: not a spell mishap just to clarify]

Disabling Phosphorus from the buildscript stopped the issue

See #414 for more information


Shibva:
Canceled: recommended to include notice for any user of AMD graphics as this may be an issue regarding phosphorus, crashing when using an AMD graphics card (you wouldn't think that this would be an issue but apparently it is because of how graphics cards in the industry are manufactured without any sort of regular norms to do things in terms of performing certain tasks;

Issue #329, opened 2022-11-25
Bloodcasting doesn't round up


yrsegal:
https://github.com/gamma-delta/HexMod/blob/main/Common/src/main/java/at/petrak/hexcasting/api/spell/casting/CastingHarness.kt#L544

This should have a `ceil` attached to it. Otherwise, small costs have trouble being bloodcast.

yrsegal:
oops, I didn't use a permalink...

https://github.com/gamma-delta/HexMod/blob/de8e84cdbd305836792ca456d95528c23c3f8c3d/Common/src/main/java/at/petrak/hexcasting/api/spell/casting/CastingHarness.kt#L545

this is the line that needed the `ceil`.

yrsegal:
This "fix" will allow free bloodcasting of any cost below half a heart, and should be reverted

yrsegal:
This still doesn't fix costs like 3.5, though. That'll just fail, even though it'll take 3 health. I said ceil for a reason

gamma-delta:
No, it doesn't. I just checked.

PR #434, opened 2023-03-08
Backports #329 fix to 1.18.2

yoyotam3:
I think this should work, I copied over the imports and fixes and changed media to mana in all the variable names.


yoyotam3:
oops no this isnt compiling what did i do wrong

yoyotam3:
Just tested it, it works and it fixed the bug.

gamma-delta:
Sure, but keep in mind I probably won't be publishing old versions (esp given that 1.20 is around the corner)

PR #451, opened 2023-04-15
Casting context

Talia-12:
Got the casting circles working, as well as fixing a number of other issues with the new casting environments, transitioning a bunch more things to use long for media amounts, etc.

Things that still need to be done: There's a *bunch* of patterns that currently flatout won't work for a spell circle (break and place blocks, grow, etc.), which really need to be fixed


Talia-12:
Also added a method on Iota to allow addon recursive data types to have max size/depth stuff work properly, and added in the negative behaviour to OpFisherman & OpFishermanButItCopies that was discussed a while back, where if the index given is negative it instead moves/copies the top element |index| deep into the stack.

PR #455, opened 2023-04-24
Update zh_cn.json

ChuijkYahus:


ChuijkYahus:
Added zh_cn.flatten.json5 to rigorously fit the en_us format.
The unflattened version would still be updated as long as the newer one is sitll WIP.

*The update is done manually, so there may be some goofy errors hidden somewhere.*

Issue #465, opened 2023-05-25
(prerelease) tooltips don't work


kristifristi:
lists, patterns and null don't show up properly

PR #468, opened 2023-06-01
Implementing Alwinfy's ArithmeticEngine

Talia-12:
The main, big change in this PR is implementing the ArithmeticEngine to allow overriding patterns based on their type signature. The mock implementation of that is [here](https://github.com/Alwinfy/ArithmeticDemo/tree/master), and I've basically just pasted it in, cleaned up the references to use the existing Hex classes for things, and added comments to everything.

I've already swapped out the systems so that what actually happens when you draw the (e.g.) Add pattern is that the Add operator for the types on top of the stack gets called, but I've left the (e.g.) Add ConstMediaAction there partly for reference so I (and you) could(can) check that everything got reimplemented properly, and partly cause deleting things feels bad.

Also in here are a fix for #465, and letting slates connect to things behind them to allow for prettier looking spell circles.


Talia-12:
okay there are now many more bug fixes also in this branch

gamma-delta:
holy shit

PR #510, opened 2023-08-07
Fix Cloth Config and PlayerBasedCastEnv unable to extract Media

Talia-12:


Common/src/main/java/at/petrak/hexcasting/api/utils/MediaHelper.kt
```diff
@@ -41,7 +41,7 @@ fun extractMedia(
  * Extract [cost] media from [holder]. If [cost] is less than zero, extract all media instead.
  * This may mutate the stack underlying [holder] (and may consume it) unless [simulate] is set.
  *
- * If [drainForBatteries] is false, this will only consider forms of media that can be used to make new batteries.
+ * If [drainForBatteries] is true, this will only consider forms of media that can be used to make new batteries.
```
object-Object, line 44: Should the comment on line 24 be updated too?

PR #511, opened 2023-08-09
Update zh_cn for 1.19

ChuijkYahus:
flatten.json5 and unflattened version would both be updated.


object-Object:
okay these changes need to go into `1.19`, but I guess it'll take a bit of Git surgery to do that

object-Object:
actually i have no idea where these need to go. I guess #512 is the 1.20 changes. @ChuijkYahus when you have a moment, could you check and see if the translations on the `1.19` branch are up to date? https://github.com/gamma-delta/HexMod/tree/1.19/Common/src/main/resources/assets/hexcasting/lang

ChuijkYahus:
Sorry for stale PR changes, the file do need to go to 1.19.
Done checking, fixed differences I could find.

object-Object:
thank you for the quick response! I just moved these changes to #535 to make it easier to merge into the 1.19 branch.

PR #512, opened 2023-08-09
Update zh_cn for 1.20

ChuijkYahus:


object-Object:
just to confirm - is this ready to merge?

ChuijkYahus:
Yes, it is. The 2 files should be the same semantically with some keys from earlier versions in `.json` file (there still could be some bugs i haven't found tho).
`.json` and `.json5` both would get updated before the file type for `en_us` is determined.

object-Object:
Actually I think only the `.flatten.json5` file is needed, not the `json` file.

object-Object:
thanks!

PR #516, opened 2023-08-22
Add region limitations to staves for bc23

LemmaEOF:
Capto_Capture 2023-08-21_11-44-50_PM

Emergency patch to add a system for area-based usage restrictions onto hexcasting wands so they're not a mass griefing tool at bc23 - documentation for the NBT values is in a comment at the top of the `AdventureHelper` file but I can write it somewhere else too if necessary

Technically untested because I cannot get this god-forsaken multiplatform environment properly set up for the life of me but it's literally just the code from yttr (used with explicit permission) and it's confirmed to work there

This is pretty critical so just get it merged by wednesday ok thanks

also includes a `.DS_Store` addition to the gitignore sorry for using a mac

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ forge*changelog.txt
 Session.vim
 plot/
 
+.DS_Store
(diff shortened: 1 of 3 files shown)


gamma-delta:
Have created a `1.20-blanketcon` branch for this because it seems more correct to me to have these blanketcon-specific patches on their own branch, rather than putting them into 1.20 and then removing them after.

Please tell me if this is a bad idea because I have no idea how we're actually getting the jar onto the server

gamma-delta:
https://github.com/gamma-delta/HexMod/tree/1.20-blanketcon

PR #518, opened 2023-08-22
Update AdventureHelper.java

walksanatora:
fix the import
(this is all i was told to fix)


object-Object:
this change is also in `Cypher121/1.20-blanketcon/more-restrictions`, so I think we can go ahead and close this.

PR #523, opened 2023-08-31
Fix wood recipe

Bawnorton:
Screenshot 2023-08-31 at 12 17 58

I don't believe this recipe is intentional


object-Object:
hey, thanks for the PR! unfortunately this file is in `.../generated/resources`, which means it's generated automatically, so we need to update the code that generated it instead.

Bawnorton:
I totally forgot about this pr and yeah I see I missed that, I can amend the pr to change the generator instead.

Bawnorton:
I can't load this project locally because of various gradle issues so I didn't have IntelliSense and couldn't run the data-gen to check but it should be correct 😅

SamsTheNerd:
looks good to me too, although we should probably run it just to make sure it works first. 

I think the edified wood variants *should* exist ? but don't currently. It looks like we don't have models for them either. Should be easy enough to add them in though, don't know if we want to do that as a part of this PR or separate, @object-Object ?

object-Object:
@SamsTheNerd depends how much effort it takes / how big of a change it would be - I see EDIFIED_WOOD in a lot of places but i'm not sure if you'd need all those for the variants?
@Bawnorton are you ok with doing that? otherwise we can just merge it and PR the variants separately.

Bawnorton:
I can't get this repo setup locally, so it would be best if someone else did that if it were done at all.

object-Object:
i feel like the ideal solution would just be to do *all* of the edified wood variant stuff in one loop and/or based on one data source somewhere (an enum?), so we don't keep forgetting to add blocks for the variants, but idk how doable that is.

object-Object:
ok - I'll run it locally, then assuming it works we can merge this and figure out the variants later.

object-Object:
Right... the reason you can't get it working locally is probably because the build is still broken on main. I'd also rather not merge into main right now since it'll probably make resolving the `1.20` merge conflicts even more of a hassle. @Bawnorton would you be able to rebase your PR onto the `1.20` branch? (or I can do it if you want)

Bawnorton:
rebased... probably better to just close this and start a new pr onto the 1.20 branch tbh lol

object-Object:
uh, yeah probably... that was not exactly the type of rebase i was talking about hah

Bawnorton:
Yeah, my bad... I just realized I rebased the wrong way... git is hard sometimes 😅

PR #526, opened 2023-09-17
Fix Media Phial consumption priority on Forge

artynova:
Priority on non-static media holder items was incorrectly hardcoded to 40 instead of the ADMediaHolder constant, 4000, so phials were being targeted last instead of first.


SamsTheNerd (commented):
seems like it will fix the issue on forge, however it still hardcodes it so that all phial-like items will have the same priority. This may be a good opportunity to expose `getConsumptionPriority()` as a part of `MediaHolderItem`, so that addons can more easily change the priority of phial-like items. Ideally it'd be defaulted to `ADMediaHolder.BATTERY_PRIORITY` so that we don't break existing items. 

@object-Object up to you if you want to merge this in for the fix. I can do the API changes once I'm home and have some time to set up the hex repo locally.

object-Object:
let's do it separately. thanks for the PR @artynova!

PR #528, opened 2023-09-24
This commit also adds edified fences as well as the sign textures 

Falkory220:


Common/src/main/java/at/petrak/hexcasting/datagen/recipe/HexplatRecipes.java
```diff
@@ -341,13 +343,28 @@ public void buildRecipes(Consumer<FinishedRecipe> recipes) {
             .pattern("WWW")
             .unlockedBy("has_item", hasItem(HexTags.Items.EDIFIED_PLANKS)).save(recipes);
 
-        ShapedRecipeBuilder.shaped(RecipeCategory.BUILDING_BLOCKS, HexBlocks.EDIFIED_STAIRS, 4)
+        ShapedRecipeBuilder.shaped(RecipeCategory.BUILDING_BLOCKS, HexBlocks.EDIFIED_STAIRS, 3)
```
object-Object: why does this give 3 now instead of 4?

Common/src/main/java/at/petrak/hexcasting/common/blocks/decoration/BlockHexFenceGate.java
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraft.world.level.BlockGetter;
+import net.minecraft.world.level.block.FenceGateBlock;
+import net.minecraft.world.level.block.state.BlockState;
+import net.minecraft.world.level.block.state.properties.WoodType;
+
+public class BlockHexFenceGate extends FenceGateBlock {
+
+    public BlockHexFenceGate(Properties $$0) {
+        super($$0, WoodType.DARK_OAK);
+    }
```
SamsTheNerd, line 16:
```java
public BlockHexFenceGate(Properties $$0) {
    super($$0, WoodType.DARK_OAK);
}
```
should probably give this a more descriptive name than $$0. I mean i guess it doesn't super matter since it's a very generic constructor but still.

Same thing on BlockHexFence.

Fabric/src/main/resources/fabric.mod.json
SamsTheNerd, line 1: is the `"paucal": "*"` right ? i assume it was changed to fix something but i would think it should have some more specific value ?

SamsTheNerd (changes requested):
looks good to me except for maybe those two comments I left.

I'm not the most familiar with hex's datagen and whatnot though so I can't say for sure that it's not missing something somewhere but it looks thorough to me !

object-Object:
let's just wait for a response from Cypher on the Blanketcon thing before we merge this.

object-Object (approved):
send it

PR #533, opened 2023-11-15
Fix OperatorUnpack predicate

Master-Bw3:
Changes the iota predicate of Vector Disintegration to check for a vector instead of a double.


SamsTheNerd (approved):
cool beans. i'm guessing it just got copied over and never changed from `OperatorPack` ?

PR #538, opened 2023-11-16
Hotfix zh_cn for 1.19

ChuijkYahus:


Common/src/main/resources/assets/hexcasting/lang/zh_cn.json
```diff
@@ -1070,7 +1070,7 @@
   "hexcasting.page.basic_spell.beep.2": "共有 16 种$(thing)乐器/$和 25 种$(thing)音高/$。两者的下标均从 0 开始。$(br2)这些乐器似乎就是$(item)音符盒/$对应的各种乐器，尽管其具体对应关系不得而知。$(br2)不管怎么说，用$(l:items/lens)$(item)探知透镜/$观察$(item)音符盒/$就能得知其乐器对应的数字。",
 
   "hexcasting.entry.blockworks": "方块工程",
-  "hexcasting.page.blockworks.place_block": "移除一个位置向量，然后挑选一个方块并放在给定位置。$(br)消耗极少量$(l:items/amethyst)$(item)紫水晶粉/$。",
```
object-Object, line 1073: why was the link removed here?

ChuijkYahus: This was to fit the original text, it makes no reference to Amethyst Dust in the first place.

The original text here:
`Remove a location from the stack, then pick a block item and place it at the given location.$(br)Costs a negligible amount of _media.`

object-Object: ahh, I see - I think I was looking at the break block entry, so I thought it referenced amethyst there. my bad, thanks!

PR #542, opened 2023-11-17
Fix wood recipe data generator

Bawnorton:
Continuation of #523 targeting the 1.20 branch

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ build
 # other
 eclipse
 run
+.DS_Store
 
 # Files from Forge MDK
 forge*changelog.txt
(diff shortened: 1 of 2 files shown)


SamsTheNerd (changes requested):
it looks like it assigns the stripped log -> stripped wood recipe twice ? once in the map loop and then once below it. 

```java
      for (var entry : EDIFIED_LOG_TO_WOOD.entrySet()) {
            var log = entry.getKey();
            var wood = entry.getValue();
            ShapedRecipeBuilder.shaped(RecipeCategory.BUILDING_BLOCKS, wood, 3)
                    .define('W', log)
                    .pattern("WW")
                    .pattern("WW")
                    .unlockedBy("has_item", hasItem(log)).save(recipes);
        }

        ShapedRecipeBuilder.shaped(RecipeCategory.BUILDING_BLOCKS, HexBlocks.STRIPPED_EDIFIED_WOOD, 3)
            .define('W', HexBlocks.STRIPPED_EDIFIED_LOG)
            .pattern("WW")
            .pattern("WW")
            .unlockedBy("has_item", hasItem(HexBlocks.STRIPPED_EDIFIED_LOG)).save(recipes);
```

also yay no .DS_Store :D

otherwise looks fine to me, as long as everything runs fine, which as far as I can tell it should ?

SamsTheNerd (approved):

PR #543, opened 2023-11-18
fix vectors with even a single 0 component being considered falsy

beholderface:
Currently, if you use Augur's Purification on a vector which has a single zero in it, it will return false. This fixes that, so it should now only return false for (0,0,0) vectors.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Vec3Iota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Vec3Iota.java
@@ -29,7 +29,7 @@ public class Vec3Iota extends Iota {
     @Override
     public boolean isTruthy() {
         var v = this.getVec3();
-        return v.x != 0.0 || v.y != 0.0 || v.z != 0.0;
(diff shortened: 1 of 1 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Vec3Iota.java
```diff
@@ -29,7 +29,7 @@ public Vec3 getVec3() {
     @Override
     public boolean isTruthy() {
         var v = this.getVec3();
-        return v.x != 0.0 && v.y != 0.0 && v.z != 0.0;
+        return v.x != 0.0 || v.y != 0.0 || v.z != 0.0;
```
object-Object:
```suggestion
        return !(v.x == 0.0 && v.y == 0.0 && v.z == 0.0);
```
I feel like this conveys the intent a bit better, right? ie. only the zero vector is falsy, everything else is truthy. (assuming i remember how Java bool syntax works)

object-Object (approved):

PR #555, opened 2023-11-29
Bring texture-based pattern rendering to 1.20

Talia-12:
Copied the code from Nitero's texture-based pattern rendering pull request into the 1.20 branch and fixed all resulting errors, then matched the colours with the currently-used colours and got the old-style rendering for slates being centered and scaled the same as the new style so that switching back to the old style when they become energised isn't jarring.


Common/src/main/java/at/petrak/hexcasting/common/command/PatternTexturesCommand.java
```diff
@@ -0,0 +1,40 @@
...
+//                                })))));
+//
+//        cmd.then(Commands.literal("textureSetResolution")
+//                .requires(dp -> dp.hasPermission(Commands.LEVEL_ADMINS))
+//                .then(Commands.argument("integer", IntegerArgumentType.integer()).executes(ctx -> {
+//                    var integer = IntegerArgumentType.getInteger(ctx, "integer");
+//                    PatternTextureManager.resolutionByBlockSize = integer;
+//                    PatternTextureManager.repaint();
+//                    return 1;
+//                })));
```
object-Object:
```suggestion
```
Probably better to just remove this - we can always grab it from commit history if we need it back

Common/src/main/java/at/petrak/hexcasting/client/render/PatternTextureManager.java
```diff
@@ -225,20 +225,28 @@ private static void vertex(Matrix4f mat, Matrix3f normal, int light, VertexConsu
     }
 
     public static ResourceLocation getTexture(List<Vec2> points, String pointsKey, int blockSize, boolean showsStrokeOrder, float lineWidth, boolean useFullSize, Color innerColor, Color outerColor) {
-        if (patternTexturesToAdd.containsKey(pointsKey))
-            return patternTexturesToAdd.get(pointsKey);
+        if (patternTexturesToAdd.containsKey(pointsKey)) {
```
object-Object, line 228: nice, i think i *tried* to do something like this but i'm bad at java and minecraft so it didn't work

PR #559, opened 2023-12-03
Fix forge crashes on death

vgskye:
Forge capabilities get nulled upon death, which is fine(I think) but causes crashes if you ever dare to die


SamsTheNerd (approved):
looks fine to me

object-Object:
thanks all!

PR #560, opened 2023-12-06
Fail Jenkins pipeline if datagen needs to be run

object-Object:
- Adds two Jenkins stages to run the datagen and fail if any files were added or modified.
- Adds a `runAllDatagen` task, which just delegates to `runXplatDatagen`, `runForgeDatagen`, and `runDatagen` (ie Fabric).
- Fixes the Patchouli/Forge version incompatibility.
- Fixes the spurious Gradle daemon error message at the end of the Forge datagen.


object-Object:
reviewed by sam, talia, aly

PR #577, opened 2024-01-09
IotaContainer interface for iota-containing iotas

Master-Bw3:
- adds an `IotaContainer` interface for iota-containing iotas. This interface is used by `getTrueNameFromDatum`, allowing for addons to implement this interface on iotas to allow for true-name write protection.


Common/src/main/java/at/petrak/hexcasting/api/casting/iota/IotaContainer.java
```diff
@@ -0,0 +1,5 @@
+package at.petrak.hexcasting.api.casting.iota;
+
+public interface IotaContainer {
```
object-Object: This should have a doc comment explaining what the interface is for - basically just what's in the PR description.

object-Object:
Actually... is there some reason we can't just use `Iota#subIotas` for this?
https://github.com/FallingColors/HexMod/blob/c065e205109b3d4683e3d84ecdaedcb8306ac824/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java#L77-L86

PR #579, opened 2024-01-11
update getTrueNameFromDatum to use Iota.subIotas()

Master-Bw3:
This allows the truename check to work on all Iotas that contain subIotas, not just lists.


object-Object:
Just double-checking - have you tested truename protection for lists ingame with this change?

Master-Bw3:
yes

object-Object:
thanks!

PR #588, opened 2024-01-16
downgrade ServerPlayer requirement to LivingEntity on CastingEnvironment

walksanatora:
also replaces places that need server player to "nullable cast" (idk the propper term) to ServerPlayer
since in all these places it is expecting a player *or* null so the logic should support caster being null 
(draft due to being untested)

Why?
would make "HexTurtle" spells more consistent(?) by allowing them all to use Minds Reflection instead of Automaton's Reflection (HexBound) or Identity Reflection (hexal). alongside this I think that each Env should have a "contract"
all getting a `CastingEnvironment` means is "we are casting, who knows if there is anything living"
getting a `CircleCastEnv` means "we are casting as a circle, still no clue if there is anything living"
geting a `LivingCastEnv` (new one being considered. if it sounds good I will most of current PlayerEnv to this) means "we are casting, there is a living being behind the cast. may not be a player". and finally `PlayerBasedCastEnv` means that "yes I am sure the caster is a human player"

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapOthersName.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapOthersName.kt
@@ -41,8 +41,9 @@ class MishapOthersName(val confidant: Player) : Mishap() {
                 val datumToCheck = poolToSearch.removeFirst()
                 if (datumToCheck is EntityIota && datumToCheck.entity is Player && datumToCheck.entity != caster)
                     return datumToCheck.entity as Player
-                if (datumToCheck is ListIota)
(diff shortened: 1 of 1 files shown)


walksanatora:
ended up somehow doing alot of broken and cursed stuff but it should *now* be ready ~~to get torn apart~~ reviewed

walksanatora:
ended up somehow doing alot of broken and cursed stuff but it should *now* be ready ~~to get torn apart~~ reviewed

walksanatora:
yeah turns out hackily doing git stuff is not good so there is stuff left over from the first iteration of the PR... and I have no clue how to resolve this

walksanatora:
still broken... I have no clue how to rebase

PR #589, opened 2024-01-21
fix weird dimensions sometimes having different great spells

beholderface:
Apparently sometimes, depending on the mods running on a server, different dimensions can be generated from different seeds, resulting in them having different great spells.
In the current version, there are a couple lines where great spell finding methods are called with a seed that isn't guaranteed to be sourced from the overworld. My PR corrects this.


Talia-12:
The change to allow SpecialHandlers access to the Environment changed what arguments `PatternRegistryManifest.matchPattern` accepts, can you rebase off of main and then change the world access inside 'PatternRegistryManifest.matchPattern' to be a `.getServer().overworld()` call?

Talia-12:
Made a separate PR to solve the issue.

Issue #561, opened 2023-12-08
Check if cleric circle player ambit is still broken


object-Object:
<img width="414" alt="image" src="https://github.com/FallingColors/HexMod/assets/37044997/18a7e13f-b019-43d5-a8c2-3a254fa34908">

Alchristis:
Still broken; the hex works inside but not outside the spell circle

![Image](https://github.com/FallingColors/HexMod/assets/59519529/a1c2e819-d6e1-426c-82b9-30ee8e6454a1)

Alchristis:
hexcasting-fabric-1.20.1-0.11.1-7-pre-609
Cleric Circle ambit works now 

hexcasting-forge-1.20.1-0.11.1-7-pre-609
Cleric Circle ambit doesn't work
To replicate, you can make the same jank spell circle as in my previous screenshot, or your own version, and see the spell cast inside but not outside of it

object-Object:
weird!

walksanatora:
this should be closed since it was fixed by #603

PR #593, opened 2024-01-23
make circles have ambit on player again

### LINKED ISSUE #591: [1.20] Bound cleric circle doesn't have ambit at bound player
**Version used: 0.11.1, build 609**

In previous versions, a bound cleric circle would always have access to the bound player along with a small radius around them. In 1.20, this does not seem to be the case. A bound cleric circle that attempts to use Mind's Reflection will fail unless the bound player is standing in the circle.

walksanatora:
closes #591


walksanatora:
also closes #561

Talia-12:
As I recall in previous versions the circle ambit around a player was PLAYER_HEIGHT rather than 32?

walksanatora:
ah right I forgot about that

Talia-12:
Ended up making a separate PR to fix the ambit range issue

PR #594, opened 2024-01-23
Added configured_feature entries for the edified trees (fixes #574)

### LINKED ISSUE #574: edifying a tree does not work
edifying a tree executes successfully but the sapling remains unchanged.

dashkal16:
Fixed edified trees by directly adding the tree configs to the common data pack.

Also fixed the resource keys being in the `minecraft` namespace as far as the code was concerned, making them impossible to find when things were registered.

I'm unhappy with this solution.  I feel like DataGen would be more appropriate, but I was unable to figure out how.


Talia-12 (approved):
Tested, confirmed fixes issue

PR #595, opened 2024-01-23
fix clerical crashes on great spells when there is no player

### LINKED ISSUE #573: 1.20 - unbound cleric impetus crash when casting great spell
When trying to use a cleric impetus to cast a great spell, if the cleric impetus is not bound to a player the game will crash once the spell circle gets to a great spell.

using hexcasting-forge-1.20.1-0.11.1-7-pre-607

[crash-2024-01-06_14.12.11-server.txt](https://github.com/FallingColors/HexMod/files/13849984/crash-2024-01-06_14.12.11-server.txt)

(setup used to reproduce:)
image

walksanatora:
fixes #573


Talia-12 (approved):
Approved

Issue #355, opened 2022-12-17
Conjuring light does not update block state


redstonekasi:
Conjuring light on a grass block that has snow on it replaces the snow but does not update the grass's state.
![image](https://user-images.githubusercontent.com/42352565/208250155-68cca8be-8366-453e-9f00-b1dcdc3ffcd2.png)

redstonekasi:
More specifically it seems like the block state does change, only for it to revert a tick later.
![output](https://user-images.githubusercontent.com/42352565/208250657-de346b82-baf1-42e7-a481-829fdbf86cce.gif)

object-Object:
What version of Hex Casting is this issue for? Do you know if this bug is still present in the 1.20.1 prereleases?

redstonekasi:
I originally found this in 1.19 Hex Casting. I tested it again in both versions just now and yes, the issue is still present.

object-Object:
Thanks! Which version exactly did you test it on? (eg. Fabric 0.11.1-7-pre-609)

redstonekasi:
The latest, build 609, yeah.

walksanatora:
fixed by #598

PR #598, opened 2024-01-23
OpConjureBlock now updates with flags 3 (updated + notify client)

dashkal16:
OpConjureBlock now updates with flags 3 (updated + notify client).
Fixes ~~#533~~ #355


Talia-12:
#533 seems to be linking to another PR, what was the issue that this is solving?

Talia-12 (approved):
Tested, fixes the issue

PR #600, opened 2024-01-23
remove dragon parts from all zone distiliations.

### LINKED ISSUE #469: Zone Distillation: Non-Living selects Ender Dragon (Multiple times)
Ender Dragon gets selected and shows up multiple times in return array, likely each hitbox segment or something. (Found while trying to blink nearby arrows. Ender Dragon got selected and even errored the spell; Can't blink Ender Dragon)

walksanatora:
fixes #469


walksanatora:
converting to draft till it is decided if we should
* remove dragon parts from all selections (current)
* remove dragon parts from Non-living (they only appear in non-item/non-player/non-monster/non-animal)
* move dragon parts from Non-Living to Living

walksanatora:
@Alwinfy has spoken
image

PR #601, opened 2024-01-26
allow special pattern handlers access to environment

walksanatora:
this opens up per-player patterns, per-pigment patterns, and alot more per-? patterns


Talia-12 (approved):
This is a simple change that allows for a number of things people want to do to be done without mixins (per-player patterns, macros, etc.), I approve!

PR #609, opened 2024-02-06
A variety of small bugfixes

### LINKED ISSUE #513: (prerelease) oplimit is way too high
getting near oplimit causes a ridiculous amount of lag/freezes the game

### LINKED ISSUE #515: Wrong Action Name
# ???
In [Phials of Media](https://gamma-delta.github.io/HexMod/#items/phials), it mentions "Archer's Purification", but it links to [Archer's Distillation](https://gamma-delta.github.io/HexMod/#patterns/basics@hexcasting:raycast).

### LINKED ISSUE #541: Expose `getConsumptionPriority()` in `MediaHolderItem`
See: https://github.com/gamma-delta/HexMod/pull/526#pullrequestreview-1734690617

### LINKED ISSUE #545: Cannot read null from empty scrolls/slates
You cannot use _Scribe's Reflection_ to read `Null` from empty scrolls or slates.

This is inconsistent with how spellbooks and foci work, as both allow players to read `Null` from their empty forms.
While scrolls and slates are not exactly the same as those other two, they are essentially the same thing when dealing with individual pattern iotas and thus should function similarly to spellbooks and foci.

Changing this would allow for easier slate writing, such as with an individual pattern copier/eraser artifact, instead of separating the artifact into an eraser and a copier.

### LINKED ISSUE #551: White Sun's Zeniths at level 0 give a very big regeneration.
White Sun's Zeniths(regeneration) at level 0 give an effect of amplifier value of -1. 
If regeneration effect's amplifier was given outside the range 0–31, remainder when it is divided by 32 is apply.
In this case, amplifier value of -1 is equrl to 31. 
Because white Sun's Zeniths at level 0 has media cost of 0, as a result player can get a very big regeneration effect with media cost of 0.

Sorry, I am not a native English speaker and this text may be ugly.

### LINKED ISSUE #585: Power Distillation should mishap when taking the square root of a negative number
Steps to reproduce:
* -1, 0.5, Power Distillation
* (-1, -1, -1), 0.5, Power Distillation

Expected result: mishap

Actual result: 0 is returned for any invalid results

Version: 0.10.3

### LINKED ISSUE #608: Advancement Triggers for Spending Media Missing
The triggers used to detect "Waste Not" and "Want Not" are not currently being called anywhere, should probably happen somewhere in PlayerBasedCastEnv so it only happens when the player is casting, but might fit better in StaffCast if those advancements where only meant to happen when staff casting (I don't remember how they worked).

Talia-12:
Fixing several issues listed below, and a number of other issues that don't (I think) have issue trackers since I noticed them as I was fixing known issues.

Fix #608, fix #585, fix #551, fix #545, fix #541, fix #515, fix #513


object-Object:
Merge conflict needs resolving, apparently

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadOffhandItem.kt
SamsTheNerd, line 1: why `!= false` ?

Common/src/main/java/at/petrak/hexcasting/common/items/storage/ItemThoughtKnot.java
SamsTheNerd, line 1: should `canWrite()` be changed to use `writeable()` instead of having a duplicate check? Actually it's prob fine for now, i imagine we'll have to rework a bunch of this when we get to 1.21 anyways.

Common/src/main/java/at/petrak/hexcasting/common/items/storage/ItemFocus.java
SamsTheNerd, line 1: should this also use `isSealed()` ?

Common/src/main/java/at/petrak/hexcasting/common/casting/arithmetic/DoubleArithmetic.kt
SamsTheNerd, line 1: doesn't look like this has a translation key?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpPlaceBlock.kt
SamsTheNerd, line 1: do we need to do `queryForMatchingStack()` twice? rather than just save it up front and make context from it after ?

SamsTheNerd (changes requested):
looks mostly fine to me. lots of file changes but they're all kinda changing a lot of the same things. Probably want to look at adding that translation key and not doing `queryForMatchingStack` twice though

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadOffhandItem.kt
SamsTheNerd: nvm, kotlin issue.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpPlaceBlock.kt
SamsTheNerd: it appears i was deceived by github code viewer into thinking those were in the same method, oops. I guess we could pass the stack into the spell but idk that it matters all that much?

SamsTheNerd (approved):
ok, fixed the stuff i wanted fixed (it was all quite small stuff)

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadOffhandItem.kt
```diff
@@ -9,23 +9,23 @@ import net.minecraft.world.InteractionHand
...
 
     override fun execute(env: CastingEnvironment, errorCtx: Context, stack: MutableList<Iota>) {
         env.mishapEnvironment.dropHeldItems()
     }
 
-    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context) = if (item.isEmpty)
+    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context) = if (item?.isEmpty != false)
         error("no_item.offhand", wanted)
     else
         error("bad_item.offhand", wanted, item.count, item.displayName)
```
object-Object: I think rearranging it like this would make the intent a lot clearer:
```suggestion
    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context) = if (item?.isEmpty == false)
        error("bad_item.offhand", wanted, item.count, item.displayName)
    else
        error("no_item.offhand", wanted)
```

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/rw/OpWrite.kt
```diff
@@ -20,17 +19,23 @@ object OpWrite : SpellAction {
     ): SpellAction.Result {
         val datum = args[0]
 
-        val (handStack, hand) = env.getHeldItemToOperateOn {
+        val (handStack) = env.getHeldItemToOperateOn {
             val datumHolder = IXplatAbstractions.INSTANCE.findDataHolder(it)
 
             datumHolder != null && datumHolder.writeIota(datum, true)
-        } ?: throw MishapBadOffhandItem.of(ItemStack.EMPTY.copy(), null, "iota.write") // TODO: hack
+        }
```
object-Object, line 26: I don't really like how duplicated this code is with OpRead, but I also don't think it's *that* big of an issue.

Common/src/main/java/at/petrak/hexcasting/common/casting/arithmetic/DoubleArithmetic.kt
```diff
@@ -57,7 +57,7 @@ object DoubleArithmetic : Arithmetic {
             MUL     -> make2 { a, b -> a * b }
             DIV     -> make2 { a, b -> if (b == 0.0) throw MishapDivideByZero.of(a, b) else a / b }
             ABS     -> make1 { a -> abs(a) }
-            POW     -> make2 { a, b -> a.pow(b) }
+            POW     -> make2 { a, b -> if (a < 0 && !DoubleIota.tolerates(floor(b), b)) throw MishapDivideByZero.of(a, b, "exponent") else a.pow(b) }
```
object-Object, line 61:
```suggestion
            // throw MishapDivideByZero if raising a negative number to a fractional power (ie. sqrt(-1) etc)
            POW     -> make2 { a, b -> if (a < 0 && !DoubleIota.tolerates(floor(b), b)) throw MishapDivideByZero.of(a, b, "exponent") else a.pow(b) }
```

Common/src/main/java/at/petrak/hexcasting/common/items/storage/ItemFocus.java
object-Object: is there any reason not to? (why was this resolved?)

object-Object (changes requested):
Mostly looks good. Added a couple suggestions for clarification.

Common/src/main/java/at/petrak/hexcasting/common/items/storage/ItemFocus.java
```diff
@@ -45,13 +44,13 @@ public String getDescriptionId(ItemStack stack) {
     }
 
     @Override
-    public @Nullable Iota emptyIota(ItemStack stack) {
-        return new NullIota();
+    public boolean writeable(ItemStack stack) {
+        return !NBTHelper.getBoolean(stack, TAG_SEALED);
```
object-Object:
```suggestion
        return !isSealed(stack);
```

object-Object (approved):
looks gamer to me

Issue #466, opened 2023-05-27
Pattern conflict checking


object-Object:
Preface: I haven't checked if this is a thing in the new registry, but I wanted to open an issue so it doesn't get lost.

There should probably be a check when patterns are added to the registry to ensure that the given angle signature doesn't already exist and intentionally crash if it does. HexxyCraft recently had an issue where I backported bool_to_number, which has the same pattern as Hexal's type/iota. The game launched fine, and the only indication was the notebook failing to load with an obscure error message which only mentioned one of the patterns involved.

gamma-delta:
Huh I swear I already impled that. Good catch

object-Object:
@Talia-12 Do you know if this was implemented on 1.20?

Talia-12:
This seems like its now handled in a way that probablyy won't have unexpected behaviour like that, but doesn't let the user know that something's gone wrong (it is being added to a Map with the pattern signature as the key, and that map is what is used to recognise patterns). I'll make it do a warning to the console if it encounters a duplicate signature.

PR #612, opened 2024-02-07
Warning when actions have same pattern

Talia-12:
Address #466 by providing a user-visible warning when two actions have patterns with the same signature (tested and having actions with the same signature no longer messes up the book).


object-Object (approved):
Looks good. Assuming you've tested it, feel free to merge.

object-Object:
send it

Issue #620, opened 2024-02-17
[1.20.1] Mod incompatibility with Fusion causing crash with Create and potentially others


karelmikie3:
First of all since this is a prerelease I understand that you don't have time to sort mod incompatibilities so it is fine if this doesn't get fixed right away. However, I have discovered an incompatibility with [Fusion](https://www.curseforge.com/minecraft/mc-mods/fusion-connected-textures) which causes a crash when combined with Create. 

Mod versions:
PAUCAL: 0.6.0-pre-118
Hex: 0.11.1-7-pre-620
Caelus API: 3.1.0+1.20
Patchouli: 1.20.1-84
Create: 0.5.1.f
Fusion: 1.1.0c
Forge: 47.2.0

Here are the logs when all mods are used:
https://gist.github.com/karelmikie3/a9c4af832fa95070b20237c2ac87d341

Now for my solution and commentary:
Looking at the log it's easy to point out `ForgeMixinDatagenModLoader`. Looking at its code shows an overwrite mixin which could be incompatible with a [similar mixin in Fusion](https://github.com/SuperMartijn642/Fusion/blob/forge-1.20/src/main/java/com/supermartijn642/fusion/mixin/DatagenModLoaderMixin.java). Changing it into an inject as done in https://github.com/FallingColors/HexMod/pull/621 alleviates this issue.

I however, have some doubts about the necessity of this mixin entirely as looking at forge's code suggests that this code would behave identically with and without the mixin.

PR #621, opened 2024-02-17
Change unnecessary mixin overwrite into an inject in ForgeMixinDatagenModLoader

karelmikie3:
This resolves https://github.com/FallingColors/HexMod/issues/620.

It changes overwriting the entire method to only injecting the additional if statement seen in
https://github.com/FallingColors/HexMod/blob/418184871b4e8f21435292ebbc3c48ab5a69634f/Forge/src/main/java/at/petrak/hexcasting/forge/mixin/ForgeMixinDatagenModLoader.java#L49-L52

This will result in a duplicate if statement like so
```java
Bootstrap.bootStrap();
ModLoader.get().gatherAndInitializeMods(ModWorkManager.syncExecutor(), ModWorkManager.parallelExecutor(), ()->{});
if (!mods.contains("forge")) {
    // If we aren't generating data for forge, automatically add forge as an existing so mods can access forge's data
    existingMods.add("forge");
}
CompletableFuture<HolderLookup.Provider> lookupProvider = CompletableFuture.supplyAsync(VanillaRegistries::createLookup, Util.backgroundExecutor());
dataGeneratorConfig = new GatherDataEvent.DataGeneratorConfig(mods, path, inputs, lookupProvider, serverGenerators,
        clientGenerators, devToolGenerators, reportsGenerator, structureValidator, flat);
if (!mods.contains("forge")) {
    // If we aren't generating data for forge, automatically add forge as an existing so mods can access forge's data
    existingMods.add("forge");
}
existingFileHelper = new ExistingFileHelper(existingPacks, existingMods, structureValidator, assetIndex, assetsDir);
```

However, this isn't a problem since `existingMods` is a Set and thus a duplicate entry won't be added.

This change makes the code changes identical to the old one while not using overwrite and thus increasing compatibility with other mods. Like fixing the crash with Fusion and Create as seen in issue https://github.com/FallingColors/HexMod/issues/620.

I do however have my doubts about this mixin's necessity. As the variable `existingMods` isn't used before the original if-statement as seen in the forge code:
https://github.com/MinecraftForge/MinecraftForge/blob/88d2dc6c8f91fd3a6b38fa06bd92cde2db1c94db/src/main/java/net/minecraftforge/data/loading/DatagenModLoader.java#L35-L55

And also isn't in the overwrite mixin:
https://github.com/FallingColors/HexMod/blob/418184871b4e8f21435292ebbc3c48ab5a69634f/Forge/src/main/java/at/petrak/hexcasting/forge/mixin/ForgeMixinDatagenModLoader.java#L39-L60

So effectively this mixin seems to change nothing with respect to the (current) forge code.


object-Object:
I agree that it doesn't seem like it would be doing anything, given that code; and it seems like the datagen runs fine with that mixin removed entirely. I think I'm just going to remove it, but if that ends up causing problems later, we can definitely revisit this PR. Thank you for the analysis/effort, and sorry for the huge delay in getting this fixed.

PR #626, opened 2024-02-27
split media env component into pre and post for finer controll

walksanatora:
todo doc and test


walksanatora:
use cases: adding external media extractions to be ran *after* the inventory has had it's media extracted as previously you could add a discovery handler but those were removed for this system (I presume) so this adds MediaExtraction Pre and Post components for Pre-env extraction and Post-env extraction modificaation

Talia-12 (approved):

PR #628, opened 2024-03-01
give the truenames their ambit back and add a config to remove it again

walksanatora:
gives truenames ambit back as it was in 1.19
and makes env.isEntityInRange and env.assertEntityInRange behave the same


Talia-12:
simplest way to do style fixes was to make a new PR from my repository, so closing this one

PR #630, opened 2024-03-02
prevent side effect throws from crashing the server

walksanatora:
This PR makes it so if a OperatorSide effect throws anything instead of crashing the server/client it will instead pass up a MishapInternalException since that is not how OperatorSideEffects should work.
Old:
currently if you throw a mishap within a RenderedSpell called via OperatorSideEffect.AttemptSpell it instantly crashes the server. this fixes that by making it try/catch the spell (if it throws a java error though it *will* still crash though... should probally fix that)


walksanatora:
okay made it also catch exceptions and wrap those into mishaps

walksanatora:
so now it catches all exceptions in all side effects and turns it into a MishapInternalException

PR #648, opened 2024-04-08
Update README.md to match gradle.properties & displayed versions

Baughn:
Well, if you look on Modrinth you'll never find a "1.x" version of Hex Casting. This updates the version numbers to match what users (and Gradle) will see.

I copied the long-term support wording. Don't know how applicable that really is.


object-Object:
thanks!

PR #653, opened 2024-04-20
Modified hexdoc renderer to have a light mode and read draw settings from CSS

JohnDog3112:
### Changes
* Made my hexdoc renderer read draw settings from CSS
* Added some cleanup code to delete previously generated images when switching renderers (This was already done when the page is closed/refreshed)
* Added overrides for the intersection points and segment renderer triangles when in light mode (made them black)

### Things to Note
* The rendered images don't automatically update when the CSS is changed as there was no event for it and I didn't want to have the client constantly checking for changes.
* Since you need an html element with the CSS styles to get their values, the only thing used to check for changes is if you selected a different palette or renderer.
  *  This means that to see changes from the CSS, you need to switch palettes/renderers

### DIFF
--- a/doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
+++ b/doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
@@ -376,7 +376,6 @@ function checkEqual(ob1, ob2) {
   return true;
 }
 
-let autoUpdateInterval = null;
 let cachedPatternImage = null;
 
 let old_settings = null;
@@ -407,19 +406,6 @@ function updateRenders() {
   }
 }
 
-function autoUpdateRenderer(interval = 500) {
-  if (autoUpdateInterval != null) {
-    clearInterval(autoUpdateInterval);
-  }
-  if (interval <= 0) {
-    console.log(`Invalid interval for autoUpdateRenderer: ${interval}`);
-    return;
-  }
-
-  autoUpdateInterval = setInterval(updateRenders, interval);
-}
-
-window.autoUpdateRenderer = autoUpdateRenderer;
 
 //palette values stored via css so they can be changedaround
 let palette_options = {
--- a/doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
+++ b/doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
@@ -319,12 +319,12 @@ let options = {
 };
 
 function load_render(name) {
-  /*if (selected == name) {
-    return;
-  }*/
 
-  if (name == "animated" && selected != name) {
-    load_animated();
+  if (name == "animated") {
+    if (selected != name) {
+      cachedPatternImage = null;
+      load_animated();
+    }
   } else {
     render_images(options[name], last_palette, name + "-settings");
   }
@@ -358,37 +358,38 @@ function checkEqual(ob1, ob2) {
 }
 
 let autoUpdateInterval = null;
-function autoUpdateRenderer(interval) {
-  let real_interval = interval;
+let cachedPatternImage = null;
+function autoUpdateRenderer(interval = 500) {
   if (autoUpdateInterval != null) {
     clearInterval(autoUpdateInterval);
   }
-
-  if (real_interval == null) {
-    real_interval = 500;
-  } else if (real_interval <= 0) {
+  if (interval <= 0) {
+    console.log(`Invalid interval for autoUpdateRenderer: ${interval}`);
     return;
   }
+
   let old_settings = null;
   autoUpdateInterval = setInterval(() => {
     if (selected == "animated") {
       return;
     }
-    let first_pattern = document.querySelector("img.spell-viz");
-    if (first_pattern == null) return;
+    if (cachedPatternImage == null) {
+      cachedPatternImage = document.querySelector("img.spell-viz");
+      if (cachedPatternImage == null) return;
+    }
 
-    let styles = getStyles(first_pattern);
+    let styles = getStyles(cachedPatternImage);
 
     let new_settings = options[selected](styles, last_palette);
     if (!checkEqual(new_settings,old_settings)) {
-      console.log("update!");
+      console.log("updated pattern render from css");
 
       old_settings = new_settings;
 
       load_render(selected);
     }
 
-  }, real_interval);
+  }, interval);
 }
 
 window.autoUpdateRenderer = autoUpdateRenderer;
@@ -403,9 +404,6 @@ let palette_options = {
 
 let last_palette = "default";
 export function load_palette(name) {
-  /*if (last_palette == name) {
-    return;
-  }*/
 
   if (!palette_options[name]) {
     return;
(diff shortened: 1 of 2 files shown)


object-Object:
For the auto-update thing, I have a couple of ideas:
- Regenerate the image if you reselect the type/palette that are already selected
- Generate the image each time the pattern drawer is opened, instead of caching it

object-Object (commented):
Other than those two comments, I think it looks good, though I'm not really that familiar with frontend stuff like this.

JohnDog3112:
> Also noticing some weird pixelation on the edges of the lines here - is there an antialiasing setting we could enable or something?
> 
> image

So, it turns out, anti-aliasing is enabled for every pattern except gradient... I'm not quite sure how that happened but it's a pretty easy fix.

JohnDog3112:
> For the auto-update thing, I have a couple of ideas:
> 
>     * Regenerate the image if you reselect the type/palette that are already selected
> 
>     * Generate the image each time the pattern drawer is opened, instead of caching it

I could do either? Though, for both, I would probably add a function to check for changes in the css when reselecting or opening a drawer. I also thought of starting a timer when it detects a manual change in the css (after doing one of the above things)

JohnDog3112:
> > Also noticing some weird pixelation on the edges of the lines here - is there an antialiasing setting we could enable or something?
> > image
> 
> So, it turns out, anti-aliasing is enabled for every pattern except gradient... I'm not quite sure how that happened but it's a pretty easy fix.

Fixed 
image

JohnDog3112:
I implemented two things for updating from css.
1. Made it so selecting same option refreshes it
2. Added a function (autoUpdateRenderer) that sets up a function to run every x ms and re-renders the images if it detects a change in the settings. It only checks the css of the first image it finds, so it won't detect changes in individual image css values, but I'm not sure why someone would do that rather than use a css class.

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -319,19 +319,80 @@ let options = {
 };
 
 function load_render(name) {
-  if (selected == name) {
+  /*if (selected == name) {
     return;
-  }
-
-  selected = name;
+  }*/
```
object-Object: this should be removed, we can always get the old code from git history if we need it

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -319,19 +319,80 @@ let options = {
...
+    real_interval = 500;
+  } else if (real_interval <= 0) {
+    return;
+  }
+  let old_settings = null;
+  autoUpdateInterval = setInterval(() => {
+    if (selected == "animated") {
+      return;
+    }
+    let first_pattern = document.querySelector("img.spell-viz");
```
object-Object: can we cache this?

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -342,9 +403,9 @@ let palette_options = {
 
 let last_palette = "default";
 export function load_palette(name) {
-  if (last_palette == name) {
+  /*if (last_palette == name) {
```
object-Object: as above

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -319,19 +319,80 @@ let options = {
...
+  /*if (selected == name) {
     return;
-  }
-
-  selected = name;
+  }*/
 
-  if (name == "animated") {
+  if (name == "animated" && selected != name) {
     load_animated();
```
object-Object: should there be a `return` here? I get this error when selecting animated twice in a row:

image

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -319,19 +319,80 @@ let options = {
...
   } else {
     render_images(options[name], last_palette, name + "-settings");
   }
+
+  selected = name;
+}
+
+function checkEqual(ob1, ob2) {
```
object-Object, line 354: huh, none of the libraries we're using have a thing for this?

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -319,19 +319,80 @@ let options = {
...
+  let real_interval = interval;
+  if (autoUpdateInterval != null) {
+    clearInterval(autoUpdateInterval);
+  }
+
+  if (real_interval == null) {
+    real_interval = 500;
+  } else if (real_interval <= 0) {
+    return;
+  }
```
object-Object: this feels confusing; why is real_interval separate from interval?

```js
interval ??= 500;
if (interval < 0) {
  console.log(`Invalid interval for autoUpdateRenderer: ${interval}`);
  return;
}
```

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
```diff
@@ -319,19 +319,80 @@ let options = {
...
+      return;
+    }
+    let first_pattern = document.querySelector("img.spell-viz");
+    if (first_pattern == null) return;
+
+    let styles = getStyles(first_pattern);
+
+    let new_settings = options[selected](styles, last_palette);
+    if (!checkEqual(new_settings,old_settings)) {
+      console.log("update!");
```
object-Object: slightly more detailed message would be nice :)

object-Object (changes requested):

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
JohnDog3112: As far as I'm aware, we aren't using any javascript libraries other than the one I made.

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
JohnDog3112: selected != name needed to be nested, otherwise, it tries to run the other renderer on the second selection.

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
JohnDog3112: I think so, I just have to add a check to clear it when switching to the animated renderer.

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
JohnDog3112: I think I was in a Rust mindset and forgot that I could mutate the parameter. I replaced it with a default parameter instead

doc/src/hexdoc_hexcasting/_templates/hexcasting_render.js.jinja
object-Object: hexdoc uses jquery and bootstrap:

https://github.com/hexdoc-dev/hexdoc/blob/b76ee52d09a1f5e829af0eb2b2f80629806a8569/src/hexdoc/_templates/index.html.jinja#L26-L38

JohnDog3112: It doesn't seem like either includes a function like that.

object-Object (approved):

PR #654, opened 2024-04-22
Fix a couple of minor issues related to the casting VM

### LINKED ISSUE #623: Evanition doesn't quite work right. It erases, but also ends the introspection
Versions
- MC:1.20.1
- Hex: Pre-619
Reproduction steps:
1. introspection
2. draw some patterns
3. Evanition
4. Bug occurs: introspection ends unexpectedly.

object-Object:
* Avoid changing the paren count when removing regular patterns with Evanition (fix #623)
  image
* Fix escapes carrying over to subsequent Thoth iterations or out of the Thoth entirely
  image
  image

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingImage.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingImage.kt
@@ -67,6 +67,11 @@ data class CastingImage private constructor(
      */
     fun withOverriddenUsedOps(count: Long) = this.copy(opsConsumed = count)
 
+    /**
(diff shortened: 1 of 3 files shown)


Talia-12 (approved):

Talia-12:
The double copy of `newImage.withResetEscape().copy(stack = tStack)` is mildly unideal, might be better to allow the withResetEscape method to accept args like `stack` so you can do `newImage.withResetEscape(stack = tStack)` and pass that through to the `copy` inside `withResetEscape` so you only copy once? no idea if that is actually possible to do

object-Object:
image

PR #670, opened 2024-05-05
[main] Add all Common sources to Fabric/Forge sources jars (fix #488)

### LINKED ISSUE #488: sources jar missing kotlin files
it looks like the sources jar for v0.10.3 in the maven doesn't have the kotlin files in it, which means vscode can't display the sources for those files. It'd be very helpful to have those in there for addon dev.

It seems like the default gradle for the fabric or architectury or whatever template I use adds the kotlin sources in by default if that helps.

object-Object:


object-Object:
image

needs more thought i guess.

PR #679, opened 2024-06-11
Add new ru_ru localization

GrinNik:


object-Object:
@GrinNik Thanks for the PR!
I did notice that there are [a couple of missing keys](https://github.com/FallingColors/HexMod/actions/runs/9468888131/job/26086811568?pr=679#step:7:33) still; those are located in [doc/resources/assets/*/lang](https://github.com/FallingColors/HexMod/tree/main/doc/resources/assets), which is a separate resource folder that is only used for the [online book](https://hexcasting.hexxy.media/v/0.11.1-7/1.0/en_us/).

GrinNik:
@object-Object Hello, should I supplement this comit or create a new one for these files?

object-Object:
Adding it to this pull request would be easiest.

PR #681, opened 2024-06-13
Update Fabric/Forge dependency files

object-Object:
* Adds a mandatory dependency on Forge for Caelus API.
* Increases the minimum version of Kotlin For Forge to 4.
* Adds a suggested dependency on Fabric for Mod Menu.


SamsTheNerd (approved):
looks fine to me. all the versions look like they match what's in gradle, some seem a little outdated, but i'm not sure that's really an issue, better to support them than to not.

PR #684, opened 2024-06-14
Anchorite's Flight forge fix

Aspurgus:


object-Object:
Actually, it seems like Fabric has the same issue, though it doesn't show up in as many cases - could you make the same change on this line while we're at it?

https://github.com/FallingColors/HexMod/blob/5294b1d7252eaafe8577611106fbb9caeaa09be9/Fabric/src/main/java/at/petrak/hexcasting/fabric/cc/CCFlight.java#L48

object-Object:
Tested on Forge and Fabric, seems to be working properly now. Thanks for the fix!

* Used Anchorite with a small radius a few hundred blocks away from the world origin
* Checked that the origin was at the expected position
* Checked that the spell stopped at the expected distance
* Checked that the spell was still active after closing and reopening the world

Aspurgus:
np !

PR #686, opened 2024-06-15
Add datagen for damage types, fix missing damage type tags

### LINKED ISSUE #622: overcasting fails if you wear armor, even unenchanted vanilla leather armor
in 1.20 (as tested on "HexxyTest"), overcasting breaks if you wear armor. Any armor. I know it's supposed to cancel the cast if the damage gets negated somehow, but it's supposed to be piercing standard attribute-based armor right?

walksanatora:
closes #622 by adding overcasting dammage to the relevant tags
(also fixes another dammage-related bug i noticed. both overcast and shame having the same id)


walksanatora:
and for once I actually tested this PR against main to make sure it did fix the issue

object-Object:
hm. actually, maybe we could datagen these?

SamsTheNerd (approved):
All the datagen stuff looks good to me. I assume the generated files all match the expected behavior for those damage types? I don't know them off the top of my head.

object-Object:
This is what was used on 1.19:

https://github.com/FallingColors/HexMod/blob/f569ba1530666f43e84584e446e6fd060395b5f4/Common/src/main/java/at/petrak/hexcasting/api/misc/DamageSourceOvercast.java#L7-L10

https://github.com/FallingColors/HexMod/blob/f569ba1530666f43e84584e446e6fd060395b5f4/Common/src/main/java/at/petrak/hexcasting/api/misc/DamageSourceShameOnYou.java#L7-L10

PR #687, opened 2024-06-18
Make flight spells cost zero if the target can already fly

beholderface:
Because I made a wisp capable of applying Anchorite's Flight on a regular basis, then realized that it would just be wasting media most of the time.
IMO the ideal solution to this would be to make the flight spells overwrite flight abilities applied by hex (but not any others), but I don't know how one would do that.


object-Object:
uh, wouldn't this just make it free to infinitely move Anchorite's around with you forever? that seems a bit broken

beholderface:
No, because it still doesn't overwrite any existing flight, and as such you'd need to exit the flight radius before it would do anything, thus making it have nonzero cost.

walksanatora:
honestly I feel like this could apply... but *only* to wayfarers
so that you could eg: deploy anchorites anywhere (and it is on you if they overlap (space or time wise)

walksanatora:
that or mabey instead of making it zero it would add time to existing wayfarers flight? (or mabey set time to that and not consume the whole ammount of media for time remaining?)

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt
```diff
@@ -35,11 +35,20 @@ class OpFlight(val type: Type) : SpellAction {
         val theArg = args.getPositiveDouble(1, argc)
         env.assertEntityInRange(target)
 
-        val cost = when (this.type) {
+        /*val cost = when (this.type) {
             Type.LimitRange -> theArg * MediaConstants.DUST_UNIT
             // A second of flight should cost 1 shard
             Type.LimitTime -> theArg * MediaConstants.SHARD_UNIT
-        }.roundToLong()
+        }.roundToLong()*/
```
object-Object: just remove it, we can get the old code from commit history if necessary

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt
```diff
@@ -35,11 +35,20 @@ class OpFlight(val type: Type) : SpellAction {
...
+        val cost = if(target.abilities.mayfly){
```
object-Object:
```suggestion
        val cost = if (target.abilities.mayfly) {
```

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt
```diff
@@ -35,11 +35,20 @@ class OpFlight(val type: Type) : SpellAction {
...
+            0L
```
object-Object, line 40: could you add a comment here either explaining why it's 0 or just linking to this PR?

object-Object (commented):
Mostly just reviewed based on code style; I'm still not 100% sure I like this solution though. It feels weird from a UX perspective to just silently fail (whether it takes any media or not).

object-Object:
If we want to make it overwrite an existing flight instead of setting the cost to 0 and silently failing, we could instead change this condition:

https://github.com/FallingColors/HexMod/blob/f1b0e8919a755e3738ee50436bae22d1193e05ec/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt#L69-L73

to something like this:

```java
  if (target.abilities.mayfly && IXplatAbstractions.INSTANCE.getFlight() == null) { 
     // Don't accidentally clobber someone else's flight 
     // TODO make this a mishap? 
     return 
 } 
```

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt
beholderface: 👍

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt
beholderface: yeah I forgot to remove the comment before opening the PR and didn't feel like bothering to make a new commit just for that

beholderface:
I've dealt with the tweaks suggested in object's review, I haven't yet done anything to do with overriding flight since it's still under discussion

gamma-delta:
I'd throw a mishap.

object-Object:
I think we decided on Discord that a good way to fix this would be to make both flight patterns mishap if the target can already fly, and also make a pattern to check if an entity can currently fly.

PR #688, opened 2024-06-20
update gradle to 8.8

Talia-12:
Update gradle version to gradle 8.8, and update all dependencies that need to be updated to match this (namely: fabric loom and ForgeGradle, plus other misc)


gradle.properties
```diff
@@ -2,14 +2,15 @@
 # This is required to provide enough memory for the Minecraft decompilation process.
 org.gradle.jvmargs=-Xmx3G
 org.gradle.daemon=false
+org.gradle.java.installations.fromEnv=JAVA_17_HOME
```
object-Object: What is this needed for?

gradle.properties
```diff
@@ -2,14 +2,15 @@
...
 
 modID=hexcasting
 modName=Hex Casting
 
 jetbrainsAnnotationsVersion=23.0.0
 
 minecraftVersion=1.20.1
-kotlinVersion=1.7.20
+kotlinVersion=1.9.0
```
object-Object: Why is Kotlin being updated? (also, if we're doing that, should KFF be updated as well?)

build.gradle
```diff
@@ -1,3 +1,9 @@
+buildscript {
+    dependencies {
+        classpath 'com.google.code.gson:gson:2.10.1'
+    }
+}
+
```
object-Object: What is this needed for?

Forge/gradle.properties
```diff
@@ -1,4 +1,4 @@
-forgeVersion=47.1.47
```
object-Object: Why is Forge being updated?

Fabric/src/main/resources/fabric.mod.json
```diff
@@ -51,10 +51,10 @@
...
     "fabric": ">=0.84",
     "fabric-language-kotlin": ">=1.9.4+kotlin.1.8.21",
-    "cardinal-components-base": "~5.2.1",
-    "cardinal-components-entity": "~5.2.1",
-    "cardinal-components-item": "~5.2.1",
-    "cardinal-components-block": "~5.2.1",
+    "cardinal-components-base": "~5.2.2",
+    "cardinal-components-entity": "~5.2.2",
+    "cardinal-components-item": "~5.2.2",
+    "cardinal-components-block": "~5.2.2",
```
object-Object: Why is CC being updated?

Fabric/gradle.properties
```diff
@@ -1,9 +1,9 @@
-fabricVersion=0.85.0+1.20.1
-fabricLoaderVersion=0.14.21
-fabricLanguageKotlinVersion=1.9.4+kotlin.1.8.21
+fabricVersion=0.92.2+1.20.1
+fabricLoaderVersion=0.15.11
+fabricLanguageKotlinVersion=1.10.9+kotlin.1.9.0
```
object-Object: Why are these being updated?

Forge/build.gradle
```diff
@@ -182,34 +177,11 @@ mixin {
     config "hexcasting_forge.mixins.json"
 }
 
-// https://discord.com/channels/313125603924639766/733055378371117127/980968358658838540
```
object-Object, line 185: yay

object-Object (commented):
need a bit of clarification

gradle.properties
Talia-12: ah, hm
that one was because that's the environment variable containing java17's install location on my machine and cradle needs to be told where it is to be able to use it, but it shouldn't break anyone else's workflow?

object-Object: can't you just set JAVA_HOME? (IntelliJ should do that for you, if configured in the project and gradle settings)

gradle.properties
Talia-12: hm
No idea why I did this one

build.gradle
Talia-12: this was a stupid bug with *something* pulling in the wrong version of gson and breaking everything, and this was how I fixed it (this was in Hexal; plausibly it wasn't going to be an issue here, I just copied all the changes I made in Hexal straight here)

Forge/gradle.properties
Talia-12: no idea, plausibly doesn't need to be

Fabric/src/main/resources/fabric.mod.json
Talia-12: no idea, plausibly doesn't need to be

Fabric/gradle.properties
Talia-12: fabric language kotlin was to match kotlin being updated, don't know for the other two

build.gradle
object-Object: Haven't seen that issue building locally; will re-add if we run into it here

object-Object (approved):
ty :)

PR #689, opened 2024-06-25
Fix the Hex Notebook not appearing in the creative tab

Auroali:


object-Object:
nice. thanks!

object-Object (approved):
Change makes sense, tested and working on Fabric.

PR #695, opened 2024-07-12
Add localization for `hexcasting.mishap.invalid_operator_args` and associated stuff

kineticneticat:
added keys for `hexcasting.mishap.invalid_operator_args` (plural & singlular)
changed `MishapInvalidIotas` to `MishapInvalidOperatorArgs` to match the class name
edited `OperationAction` to account for above

image


Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
```diff
@@ -0,0 +1,54 @@
...
+    private fun collateIotas(iotas: List<Iota>): MutableComponent {
+        val out = MutableComponent.create(ComponentContents.EMPTY)
+        for (i in iotas.indices) {
+            out.append(iotas[i].display())
+            if (i < iotas.size-1) {
+                out.append(", ")
+            }
+        }
+        return out
+    }
```
object-Object: I think this could be simplified to something like this:
```kt
ComponentUtils.formatList(iotas.map { it.display() }, ", ".asTextComponent)
```
`ComponentUtils.formatList` seems to be the component equivalent of `join`.

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.network.chat.ComponentContents
+import net.minecraft.network.chat.MutableComponent
+import net.minecraft.world.item.DyeColor
+
+/**
+ * The value failed some kind of predicate.
+ */
+class MishapInvalidOperatorArgs(
+    private val perpetrators: List<Iota>
+) : Mishap() {
```
object-Object: nit: I think this would look fine on a single line now

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
```diff
@@ -0,0 +1,54 @@
...
+import net.minecraft.network.chat.Component
+import net.minecraft.network.chat.ComponentContents
+import net.minecraft.network.chat.MutableComponent
+import net.minecraft.world.item.DyeColor
+
+/**
+ * The value failed some kind of predicate.
+ */
+class MishapInvalidOperatorArgs(
+    private val perpetrators: List<Iota>
```
object-Object: does the visibility here need to be changed to private?

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
```diff
@@ -0,0 +1,54 @@
...
+            error(
+                "invalid_operator_args.plural",
+                perpetrators.size,
+                0,
+                perpetrators.lastIndex,
+                collateIotas(perpetrators)
+            )
+        }
+    }
+    private fun collateIotas(iotas: List<Iota>): MutableComponent {
```
object-Object: nit:
```suggestion
    }

    private fun collateIotas(iotas: List<Iota>): MutableComponent {
```

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
```diff
@@ -0,0 +1,54 @@
...
+    override fun accentColor(ctx: CastingEnvironment, errorCtx: Context): FrozenPigment =
+        dyeColor(DyeColor.GRAY)
+
+    override fun execute(env: CastingEnvironment, errorCtx: Context, stack: MutableList<Iota>) {
+        for (i in perpetrators.indices) {
+            stack[stack.size - 1 - i] = GarbageIota()
+        }
+    }
+
+    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context): Component {
```
object-Object, line 25: does this behave reasonably well if perpetrators is empty?

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -890,6 +890,11 @@
         cos: "the cosine of %s",
       },
 
+      invalid_operator_args: {
+        single: "Got an unexpected iota at index %d of the stack: %s",
+        plural: "Got %d unexpected iotas at indices %d-%d of the stack: %s",
```
object-Object: nit: I think `one` and `many` would be more similar to other existing keys; not by any means a huge issue though, up to you

object-Object: nit: the leading case is inconsistent across mishaps, but `invalid_value` is lowercase, so I think it makes sense to match that

```suggestion
        single: "got an unexpected iota at index %d of the stack: %s",
        plural: "got %d unexpected iotas at indices %d-%d of the stack: %s",
```

object-Object (commented):
overall looks good; I left a few comments, but none of them are *really* a big deal. up to you which ones you want to address. tested it locally, seems to work fine

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
kineticneticat: removed `collateIotas` due to above replacement

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidOperatorArgs.kt
kineticneticat: it returns "got 0 unexpected iotas at indices 0--1 of the stack:" which seems fine as you *should* never see this

object-Object (approved):
Looks great - thanks for the fix!

PR #697, opened 2024-07-13
Turn Environment creation event into a BiConsumer to also take userData

walksanatora:
6 lines to get this working...

why: would allow persisting modifications to env for example casting env components or (the main reason why I am making this) hexdim and everett's environment


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -42,21 +43,21 @@ public abstract class CastingEnvironment {
     /**
      * Stores all listeners that should be notified whenever a CastingEnvironment is initialised.
      */
-    private static final List<Consumer<CastingEnvironment>> createEventListeners = new ArrayList<>();
+    private static final List<BiConsumer<CastingEnvironment, CompoundTag>> createEventListeners = new ArrayList<>();
 
     /**
      * Add a listener that will be called whenever a new CastingEnvironment is created.
      */
-    public static void addCreateEventListener(Consumer<CastingEnvironment> listener) {
```
object-Object, line 51: Would it be worth adding an overload for the old signature? something like `createEventListeners.add((env, userData) ->listener.invoke(env))` (syntax??)

object-Object:
Alwinfy's comments:
image
image

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -43,21 +45,28 @@ public abstract class CastingEnvironment {
...
+    public static void addCreateEventListener(BiConsumer<CastingEnvironment, CompoundTag> listener) {
         createEventListeners.add(listener);
     }
 
+    /**
+     * Add a listener that will be called whenever a new CastingEnvironment is created (legacy).
+     */
+    public static void addCreateEventListener(Consumer<CastingEnvironment> listener) {
```
object-Object, line 62: Should probably add a deprecated annotation to this function.

PR #701, opened 2024-07-17
Changed `GaslightingTracker` to be instanced, not static

kineticneticat:
Made `GaslightingTracker` able to be instanced, allowing multiple different gaslighting timers on separate predicates
Obviously this isn't too useful in base hexcasting right now as there's only the quenched allay group, but the new class is in api so that it is accessible for addons & the like

There's probably better ways to do some of these things, so any pointers would be great :)
image


Common/src/main/java/at/petrak/hexcasting/api/client/GaslightingTracker.java
SamsTheNerd, line 1: the `GASLIGHTING_TRACKERS` field should be private and values should be accessed through some static getter so that the internal implementation is not a part of the api. 

the fields which are now nonstatic should be `camelCase` 

Why `LinkedHashMap` rather than just `HashMap` ? 

`gaslightingPred` should be final.

Common/src/main/java/at/petrak/hexcasting/client/ClientTickCounter.java
SamsTheNerd, line 1: this file should call a static method on `GaslightingTracker` that handles updating the trackers rather than iterating through them out here.

Common/src/main/java/at/petrak/hexcasting/common/lib/hex/HexGaslighting.java
SamsTheNerd, line 1: don't need this class imo, just put the static field in `GaslightingTracker` and it'll get initialized whenever that class gets referenced (which will happen when it's first used, so no need to init it)

Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexInitializer.kt
SamsTheNerd, line 1: same as previous, don't need this init

SamsTheNerd: also, for future reference this should probably have gone in `RegisterClientStuff` (and honestly that class should probably be renamed to make it more findable but whatever, future problem) so that it's in common and only on the client

Forge/src/main/java/at/petrak/hexcasting/forge/datagen/xplat/HexItemModels.java
SamsTheNerd, line 1: I don't think that method needs to take in a predicate if they all use the same one. or should atleast have a second method that doesn't take in the predicate and just delegates to the one that does. 

Also, it seems redundant to take in a resLoc, use that to get the tracker, and then get the resLoc from it? Taking in the `GaslightingTracker` itself (since it's a static field) seems like the neatest/most readable imo.

I'm not sure we need to change the actual predicate key though? there's not really a case where an item would have multiple gaslight predicates so I think just keeping it as `hexcasting:variant` is fine. If we did this I'm not sure that these would even need to take in anything or be changed much at all?

SamsTheNerd: (nitpick) also be mindful of consistent tabs and newlines, etc

Common/src/main/java/at/petrak/hexcasting/client/RegisterClientStuff.java
SamsTheNerd, line 1: similar thoughts to the datagen stuff, not a fan of having the methods take in strings when they're all the same.

I guess i didn't give reasons for that, it can lead to easily missed typos, makes it harder to change later, and if it's just a string literal rather than referencing one consistent object/field (like the old static `GaslightingTracker.GASLIGHTING_PREDICATE`) then it can lead to changing it in one place and forgetting to change it in every other place, rather than just changing it in the central place and everything else changing automatically since it just references it.

also, should probably not call the resLocs predicates when you're using them to refer to the gaslighting tracker id if we're not changing the predicate (see datagen comments).

SamsTheNerd (changes requested):
Overall looks like a good start ! needs a bit of cleanup though, and I'd like to get another opinion on whether or not to change the model predicate/how to organize datagen for it, @object-Object if you have opinions.

object-Object:
Just a heads up: we likely will not be merging this or putting more effort into it until after we get a 1.20 release out. It seems to me like a quite specific edge case feature that base Hex doesn't really have a use case for, and the amount of changes makes me wary. I don't have any reason to be against the change, it just shouldn't be a priority right now. ^^

kineticneticat:
Yea that's absolutely fair, I just made this pr since I had gotten it to work in my addon, so I figured I might as well put it here since it might be useful if a separate gaslighting group was needed at some point.

I'll at least get those edits done so it's in a better state if/when it's needed!

Forge/src/main/java/at/petrak/hexcasting/forge/datagen/xplat/HexItemModels.java
kineticneticat: I don't know if it'll work for you, but what I've done is have the `GaslightingTracker`s be stored with separate keys so that they can have separate timers and cooldowns, but then they all use the same key for the model shenanigans since as you said, there's not really a case for having two.

PR #702, opened 2024-07-22
Update ru translation.

LedinecMing:
This translates most of english terms, fixed spelling mistakes.
Naming is just like in english based on amount of in and out arguments.
All hexes are translated.
The last untranslated piece is text of lore fragments.


object-Object:
Thank you!

PR #706, opened 2024-07-23
Remove ForgeMixinDatagenModLoader to fix conflict with Create/Fusion/others

### LINKED ISSUE #620: [1.20.1] Mod incompatibility with Fusion causing crash with Create and potentially others

object-Object:
Fixes #620.

Supersedes #621 - it seems like datagen and the mod itself both work fine without this mixin being present at all. I've confirmed that the crash with Fusion does not occur with this change.


SamsTheNerd (approved):
seems fine. 

summary from DMs for future reference: nobody knows what this mixin was for, issues go away when it's gone and everything works. If we find out we need it in the future gate it with a mixin plugin.

PR #708, opened 2024-08-01
Inline Patterns !! (and also a small little pattern rendering rework)

### LINKED ISSUE #587: [1.20] Key-pattern displayed on Akashic Bookshelf isn't lit correctly
**Version used: 0.11.1, build 609**

The key-pattern rendered on the front of a filled Akashic Bookshelf appears nearly pitch black rather than the normal white, as though it's not receiving any lighting.

SamsTheNerd:
This adds [inline](https://github.com/SamsTheNerd/inline) as a dependency and changes the pattern iota display to render the pattern visually rather than `HexPattern(dir, angles)`. It also adds some Inline bits to entity iotas (currently only player heads are enabled because inline's serialization seems mildly broken but that's a future issue !)

Instead of re-implementing/copying all of the pattern rendering bits that i'd need for this though, i spent some time cleaning up and re-organizing the existing pattern rendering system. Although this is more of an additional system that helps with pattern rendering rather than a rewrite of the core RenderLib code. You can read the details of it [here](https://github.com/FallingColors/HexMod/blob/inlinepatterns/Common/src/main/java/at/petrak/hexcasting/client/render/PATTERN_RENDER_LORE.md).

Some benefits of this new system:
- configuring pattern rendering through clear and reuseable settings
- less duplicate code, especially for pattern positioning
- texture and dynamic (wobbly/gradient/renderlib one) renderers are much more in sync with eachother
  - share more code
  - almost identical outputs
  - abstracted behind `PatternRenderer` so you don't have to Choose which one to use, or do anything different to use the other, besides changing settings.

Some other fun changes:
- slates glow amethyst-purple now !
- texture rendering is split into 2 textures (inner and outer) so that textures can be re-used with different colors
- dynamic renderer can use other shaders, formats, and vertexconsumers (this means light and normals !!)
- scrolls & slates show their pattern in their name for better hotbar clarity.

Some images:

image
image
image
image
image

Texture renderer vs dynamic (they are almost identical lol):
image
image


SamsTheNerd:
next inline version has some breaking changes i'd like to push in before this gets merged. can still be reviewed, all the big pattern rendering things aren't gonna change.

SamsTheNerd:
see #764

PR #714, opened 2024-08-14
Make Break Block check for permission

### LINKED ISSUE #678: Break Block does not check for breaking permission, unlike other blockworks spells
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-637

### Issue description

In the 1.19 version of the mod, the Break Block spell calls `IXplatAbstractions.INSTANCE.isBreakingAllowed` to ensure that you can't break blocks in other people's claims. However, in the 1.20 version, that check is no longer there, so there's nothing preventing you from using Break Block inside someone else's claim. This is especially odd since all of the other spells which remove blocks (namely Edify Sapling, Extinguish, and Destroy Liquid) still have the `isBreakingAllowed` check and thus still respect claims.

### Other information

This may be the root cause for some of the effects mentioned in #450 - specifically, the fact that in 1.20 you can break blocks in other people's claims despite most of the other blockworks spells (correctly) failing to work in claimed territory.

vgskye:
Fixes #678.


object-Object:
Confirmed that the issue is fixed by attempting to break a block in a claim created by my alt. The block broke before this fix and does not break now. Thanks for the PR!

PR #718, opened 2024-08-17
Add Codecs to serializable classes

Master-Bw3:
adds codecs to serializable classes like Iotas, Continuation Frames, HexPattern, etc. Although codecs are not necessary for 1.20.1, this should help us migrate to future versions more easily.

## Changes

All `serialize()` and `deserialize()` methods in classes with codecs have been marked as deprecated as they should be removed in future versions. The implementations of those methods have also been altered to use codecs.

2 new extension functions have been added to `HexUtils.kt` to aid with using codecs:
```<T : Any> T.serializeWithCodec(codec: Codec<T>): Tag?```
```<T : Any> Tag.deserializeWithCodec(codec: Codec<T>): T?```

All codecs were designed to be backwards compatible, so iotas from previous versions should load correctly.

Entity Iotas have been changed to accommodate codecs:
- Entity iotas no longer hold a direct reference to an entity
- `List<Iota>.getEntity` and related functions now require a `world` argument
- When an entity iota is deserialized, if the entity can't be found in the world, it will instead deserialize to a Null iota. If the entity is a player, then the iota will be marked as a truename.


walksanatora:
q: why do the singleton iotas (null/garbage) go through codec deserialization when they are Codec.unit's cant you just safely assume that the iota is just the unit

gamma-delta:
Can you rebase this into fewer commits please? Like one commit for the codec implementation, and one commit where you switch everything over to using the new codecs.

gamma-delta:
I think we should probably delay this until after the 1.20.1 release. @object-Object I think you had similar thoughts?

PR #719, opened 2024-08-19
Allow for playerless block raycasts

### LINKED ISSUE #668: 1.20.1 unbound circles unable to cast archers distillation
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-634

### Modloader version

Fabric Loader 0.15.9 + Fabric API 0.92.1+1.20.1

### Modpack info

hexxycraft 1.20.1 pack

### The latest.log file

_No response_

### Issue description

unbound cleric impetus circles are unable to perform raycasts with archers distillation
image

### Steps to reproduce

_No response_

### Other information

_No response_

vgskye:
fixes #668.
this was made in Codespaces, so this is untested code; it *should* work but testing may be advisable


object-Object:
huh.

https://github.com/FallingColors/HexMod/blob/a763cdfbd78d12a3069b1b4a982a08793a48b162/Fabric/src/main/java/at/petrak/hexcasting/fabric/mixin/FabricClipContextMixin.java#L14-L23

object-Object:
So it turns out this was already fixed, actually. Talia added an equivalent mixin last year in 717dd3e5. This issue was introduced in 1a44b31a; the actual bug is just the unchecked cast in OpBlockRaycast. It works fine on both Forge and Fabric if that cast is removed. I guess Forge does something like this mixin already? Not sure why it's not needed on Forge.

SamsTheNerd (approved):
joyous

PR #723, opened 2024-08-21
Fix #672 - Invalid-inputs mishap from Thoth reads the stack backwards

### LINKED ISSUE #672: Invalid-inputs mishap from Thoth reads the stack backwards
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-634

### Modloader version

Fabric: loader 0.15.9, api 0.92.1+1.20.1

### Modpack info

HexxyTest (https://github.com/walksanatora/HexxyTestPack)

### The latest.log file

_No response_

### Issue description

When you attempt to cast a pattern without having the correct input types on the stack, you'll get a mishap message saying `<spell>: expected a <type> at index <number> of the stack, but got <actual value>`. Normally, the index in this message counts down from the top of the stack – ie, if the topmost iota is the problem it'll say "index 0" and if the second iota down is the problem it'll say "index 1". 

However, when the pattern triggering the mishap is Thoth's Gambit, it acts differently. It seems to count the indices from the _bottom_ of the stack instead of the top, with the lowest iota being "index 0" and the top two (which are the only ones thoth actually cares about) being "index N-1" and "index N-2" with N as the current size of the stack.

Even stranger, the actual effect of the mishap (that is, replacing the problematic iota with Garbage) seems to be reading this incorrect number according to the normal system. If Thoth's Gambit gets an incorrect type at the top of a stack of 5 iotas, the mishap message will read "index 4" and then replace the iota which is _actually_ at index 4 (the one all the way at the bottom) with Garbage.

### Steps to reproduce

1. Push any number of iotas onto the stack, making sure that at least one of the top two is not a valid input for Thoth's Gambit
2. Cast Thoth's Gambit, and observe that an iota all the way at the bottom of the stack gets Garbage'd instead of one of the top two
3. Cast any other pattern for which the top iotas are not valid, and observe that the correct iota gets Garbage'd

### Other information

A full video demonstration of the bug can be found here: https://imgur.com/a/wHALWxN
image

navarchus:
Fix https://github.com/FallingColors/HexMod/issues/672

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
@@ -17,10 +17,8 @@ object OpForEach : Action {
         if (stack.size < 2)
             throw MishapNotEnoughArgs(2, stack.size)
 
-
         val instrs = stack.getList(stack.lastIndex - 1, stack.size)
         val datums = stack.getList(stack.lastIndex, stack.size)
-
         stack.removeLastOrNull()
         stack.removeLastOrNull()
 
--- a/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
@@ -1,5 +1,6 @@
 package at.petrak.hexcasting.common.casting.actions.eval
 
+import at.petrak.hexcasting.api.casting.SpellList
 import at.petrak.hexcasting.api.casting.castables.Action
 import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
 import at.petrak.hexcasting.api.casting.eval.OperationResult
@@ -7,6 +8,8 @@ import at.petrak.hexcasting.api.casting.eval.vm.CastingImage
 import at.petrak.hexcasting.api.casting.eval.vm.FrameForEach
 import at.petrak.hexcasting.api.casting.eval.vm.SpellContinuation
 import at.petrak.hexcasting.api.casting.getList
+import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.casting.mishaps.MishapInvalidIota
 import at.petrak.hexcasting.api.casting.mishaps.MishapNotEnoughArgs
 import at.petrak.hexcasting.common.lib.hex.HexEvalSounds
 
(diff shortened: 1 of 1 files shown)


SamsTheNerd (approved):
rad

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
```diff
@@ -1,12 +1,15 @@
...
+import at.petrak.hexcasting.api.casting.SpellList
 import at.petrak.hexcasting.api.casting.castables.Action
 import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
 import at.petrak.hexcasting.api.casting.eval.OperationResult
 import at.petrak.hexcasting.api.casting.eval.vm.CastingImage
 import at.petrak.hexcasting.api.casting.eval.vm.FrameForEach
 import at.petrak.hexcasting.api.casting.eval.vm.SpellContinuation
 import at.petrak.hexcasting.api.casting.getList
+import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.casting.mishaps.MishapInvalidIota
```
vgskye: vestigial imports?

SamsTheNerd: oh true !

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
```diff
@@ -1,12 +1,15 @@
...
 
+import at.petrak.hexcasting.api.casting.SpellList
 import at.petrak.hexcasting.api.casting.castables.Action
 import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
 import at.petrak.hexcasting.api.casting.eval.OperationResult
 import at.petrak.hexcasting.api.casting.eval.vm.CastingImage
 import at.petrak.hexcasting.api.casting.eval.vm.FrameForEach
 import at.petrak.hexcasting.api.casting.eval.vm.SpellContinuation
 import at.petrak.hexcasting.api.casting.getList
+import at.petrak.hexcasting.api.casting.iota.Iota
```
SamsTheNerd: what skye said, unnecessary imports, those should prob be removed.

SamsTheNerd (changes requested):

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
navarchus: removed!

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/eval/OpForEach.kt
```diff
@@ -17,14 +17,15 @@ object OpForEach : Action {
...
+
+        val instrs = stack.getList(stack.lastIndex - 1, stack.size)
+        val datums = stack.getList(stack.lastIndex, stack.size)
+
         stack.removeLastOrNull()
         stack.removeLastOrNull()
 
         val frame = FrameForEach(datums, instrs, null, mutableListOf())
         val image2 = image.withUsedOp().copy(stack = stack)
-
```
vgskye: nit: weird whitespace change. won't even show up in blame since deleted line so it doesn't matter but still

SamsTheNerd (approved):

PR #725, opened 2024-08-21
Fix sounds not fading with distance by changing stereo sound files to mono

Shibva:
Edited certain sound files that are Stereo-channeled to be instead be mono-channeled due to Minecraft only having Volume Attenuation for .ogg's that are mono-channeled; thus causing stereo ones to play as if they were global (it is in fact how Mojang introduced those sounds too to be global)

-This effectively corrects any issues in regards to hexcasting sounds not fading in volume with distance and will stay that way as long as future/updated sounds remain **MONO-CHANNELED** (unless of course, you want it to be global = ) )

- Replaced audio files have also been trimmed to reduce filesize even if slightly.
   - Effort was made to keep the sounds as original as possible, and went over it with a fine-tuned hammer even
   
- Shoutout to the Figura Helper team for helping me with a personal project that leads to me discovering the root cause of this issue
   - Also to "Jess, the confused cube eater" for suggesting to check for stereo sound files in base hex as well; nice catch
   
[this PR merge request was made at the request of Walksanator =D]


SamsTheNerd:
were the other casting sounds already mono?

Shibva:
Yep, I combed thought all of them and these are the only ones that have stereo instead of mono

Also feel free to also make adjustments to the sound events now that sound fadeoff is working, might need a few tweaks. 

Sadly that part is outside my capabilities 

Thanks for asking

PR #727, opened 2024-08-22
Exclude loreInjections from auto config to have config work

### LINKED ISSUE #399: [0.11.0-pre-548] Unable to change mod configs through Mod Menu
Using Fabric.
image
(no config button in the top right)

vgskye:
fixes #399. or should `loreInjections` be turned into a `List<String>` so it can be configured via mod menu?


vgskye:
...thinking about it I think #728 is the better solution

PR #728, opened 2024-08-22
Turn loreInjections and DEFAULT_LORE_INJECTS into strings for Auto Config

### LINKED ISSUE #399: [0.11.0-pre-548] Unable to change mod configs through Mod Menu

vgskye:
also fixes #399. I'm not sure if this or #727 would be the better approach so I did both


object-Object:
I'm thinking maybe it would be better to basically copy what `scrollInjections` does? Then we could still use ResourceLocation instead of raw strings.

Issue #652, opened 2024-04-20
Add OpThanos to the pattern registry


object-Object:
As of 488c6feac, it seems like OpThanos, despite being implemented, is not actually used anywhere.
![image](https://github.com/FallingColors/HexMod/assets/37044997/77bbf5b8-1258-42d7-b648-07f705e24fba)

walksanatora:
in 1.19 this was the "meta evals left" pattern. what was the pattern for it back then? (or has it remained unused for it's entire lifetime?)

PR #731, opened 2024-08-23
Implement thanatos' reflection (placeholder name for OpThanos)

navarchus:
Address #652

### DIFF
--- a/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
+++ b/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
@@ -770,7 +770,8 @@
         "eval/cc": "Iris' Gambit",
         for_each: "Thoth's Gambit",
         halt: "Charon's Gambit",
-        
+        thanatos: "Thanatos' Reflection",
+
         "interop/": {
           "gravity/": {
             get: "Gravitational Purification",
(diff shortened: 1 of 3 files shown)


Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -1778,6 +1779,8 @@
 
         "eval/cc.1": "Cast a pattern or list of patterns from the stack exactly like $(l:patterns/meta#hexcasting:eval)$(action)Hermes' Gambit/$, except that a unique \"Jump\" iota is pushed to the stack beforehand. ",
         "eval/cc.2": "When the \"Jump\"-iota is executed, it'll skip the rest of the patterns and jump directly to the end of the pattern list.$(p)While this may seem redundant given $(l:patterns/meta#hexcasting:halt)$(action)Charon's Gambit/$ exists, this allows you to exit $(italic)nested/$ $(l:patterns/meta#hexcasting:eval)$(action)Hermes'/$ invocations in a controlled way, where Charon only allows you to exit one.$(p)The \"Jump\" iota will apparently stay on the stack even after execution is finished... better not think about the implications of that.",
+
+        "thanatos.1": "Adds the number of evaluations a _Hex is still capable of to the stack. This is reduced by $(l:patterns/meta#hexcasting:eval)$(action)Hermes'/$ and $(l:patterns/meta#hexcasting:for_each)$(action)Thoth's/$ Gambit."
```
object-Object: Maybe something like this instead? Thanatos in 1.20 is different than in 1.19, since we replaced the eval limit (increased by Hermes and Thoth) with the op limit (increased by each pattern executed).

```suggestion
        "thanatos.1": "Adds the number of patterns a _Hex is still capable of evaluating to the stack. This is reduced by one for each pattern cast by the _Hex."
```

object-Object:
thanks!

Issue #649, opened 2024-04-12
mishap.location_to_far localization key, among others, might be missing


Twisted-Code:
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-620

### Modloader version

Fabric: 0.15.9 + API 0.91.0+ 1.20.1

### Modpack info

https://github.com/walksanatora/HexxyTestPack

### The latest.log file

https://mclo.gs/UxqbiMk

### Issue description

I'm not sure if this happens for ALL keys but I know that I've seen this particular key get localized properly when I cast by hand. More testing may be required though, I just wanted to get the ball rolling!

### Steps to reproduce

1. Set up a Spell Circle as you usually would
2. Install a hex that will cause location_too_far (Or possibly others)
3. activate the circle
4. Issue occurs, at least in my current context...

### Other information

Screenshot:
![2024-04-11_20 14 30](https://github.com/FallingColors/HexMod/assets/7110033/a1279ce7-2a0d-4b78-b604-6e9de2f76b71)

macks2008:
![image](https://github.com/FallingColors/HexMod/assets/7110033/4d982f8a-21d3-4d66-859a-7257b4859132)
Okay I might be wrong about this being something with the impetus. Is the key just missing, because if so that's an easy enough fix. I don't think it's the only missing key though, so maybe you want to have a branch for all related fixes to go into before merging into main?
Let me know what you want to do, but I think I can handle this if it turns out it's a missing key. That's trivial to fix.

object-Object:
I'm fine with small PRs, if you do want to fix it.

PR #733, opened 2024-08-23
add localization entry for location_too_far

navarchus:
address #649 (Tested, works on fabric)


object-Object:
oh, huh.

https://github.com/FallingColors/HexMod/blob/b4aa31996bd0983135e4b949f30d2f24f4fd2f04/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5#L958-L964

https://github.com/FallingColors/HexMod/blob/b4aa31996bd0983135e4b949f30d2f24f4fd2f04/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadLocation.kt#L19-L21

So the actual bug is that the keys in the file are eg. `hexcasting.mishap.location.too_far` (with a period), but the mishap is constructing keys like `hexcasting.mishap.location_too_far` (with an underscore).

On 1.19, these keys used an underscore, so I think the best fix would be to update the file to match.

We should be able to do that just by changing this line from `location: {` to `"location_": {` .

https://github.com/FallingColors/HexMod/blob/b4aa31996bd0983135e4b949f30d2f24f4fd2f04/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5#L958

navarchus:
yep that works lol

object-Object:
cool cool
actually, I guess we may as well also update the other lang files (zh_cn, ru_ru) with the same change if you can? I should have mentioned it before, sorry

object-Object:
lmfao thanks

PR #734, opened 2024-08-23
Allow spell circles to access player inventory

### LINKED ISSUE #638: Spell Circle Unable to use Recharge
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-629

### Modloader version

Forge 47.2.0

### Modpack info

Custom made pack.

### The latest.log file

https://mclo.gs/B2HZ1IO

### Issue description

Spell circles are unable to use the recharge hex to recharge a phial in a players offhand.

### Steps to reproduce

Create a Spell circle that grabs amethyst in its bounds in zone distillation, and uses thoth's to cast recharge with each of them. It will fail saying a rechargeable item is needed in your other hand, even if you are holding an empty phial (no matter which hand or whats in the other hand).

### Other information

_No response_

vgskye:
fixes #638

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
@@ -19,6 +19,7 @@ import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.phys.Vec3;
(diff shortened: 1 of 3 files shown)


SamsTheNerd:
this code looks basically copy pasted from the player casting env, maybe it should just be abstracted into a shared method somewhere?

vgskye:
it is based off of that, but there are some behavioural changes(e.g. pull mainhand first then offhand, rather than other-hand-first or other-hand-only since spell circles are hands-free) that I'm not sure makes sense to unify into one set of methods

SamsTheNerd (changes requested):
i think pulling it out and unifying still makes the most sense. can have it just take a list of hands or something and that should handle that behavior? That seemed like the main thing, was there another issue with unifying it ?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -360,11 +361,88 @@ public InteractionHand getOtherHand() {
...
+                    var mainhand = caster.getItemInHand(InteractionHand.MAIN_HAND);
+                    if (!mainhand.isEmpty()) {
+                        out.add(mainhand);
+                    }
+
+                    var offhand = caster.getItemInHand(InteractionHand.OFF_HAND);
+                    if (!offhand.isEmpty()) {
+                        out.add(offhand);
+                    }
+                } else {
```
SamsTheNerd, line 379: shouldn't this case check both hands, just prioritizing otherhand. for stuff like self refilling artifacts and whatnot.

SamsTheNerd: i think having this method take a list of hands to check may be better? that way callers can prioritize whichever hand they want while also being able to decide if they want the other hand used also. then just loop over that list, would probably be neater too.

SamsTheNerd: not sure that this would ever actually be used in a case like that though

SamsTheNerd (commented):
other than comment I left (which I'm not sure about), code looks fine. Should probably be tested a bit more before merging but I'm not doing that rn

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
vgskye: iiuc this is used for Place Block&co, not Recharge&co? I was just retaining previous behaviour when provided with a casting hand though

SamsTheNerd (approved):

Issue #729, opened 2024-08-22
Missing translation keys for config options


vgskye:
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-659

### Issue description

Currently masked by #399, but it seems like there are these translation keys needed:
(prefix of `text.autoconfig.hexcasting.option.server.` omitted, all also need their tooltips with `.@Tooltip` suffixed)
- `doesTrueNameHaveAmbit`
- `tpDimDenylist`
- `loreInjections`
- `loreChance`

object-Object:
Seems the lore keys were added in #838, so this will be fixed by #921.

PR #735, opened 2024-08-23
Add option to have click toggle drawing instead of dragging

### LINKED ISSUE #357: client config option Click type option ->"click (start) then click(end)" rather than "click hold(start + continue) then release(end)" client config option?
is there a plan ever to add a local config to do spells via "click (start) then click(end)" rather than "click hold(start + continue)  then release(end)"

For accessibility for those who don't want to hold click for patterns (especially longer ones) & release click still there for those who want less clicks.

vgskye:
closes #357. confirmed working on fabric, forge should probably work


vgskye:
tested and working on forge as well

object-Object:
We should probably also add a localization to the lang file for this cloth config entry. see here:

https://github.com/FallingColors/HexMod/blob/f3f9d0740bda6df22826c18780e0241500983e6b/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5#L286-L303

vgskye:
> We should probably also add a localization to the lang file for this cloth config entry.
I feel like this can be batched with #729?

object-Object:
i don't see any reason to? just because other options didn't get localizations doesn't mean we should continue not adding them for new ones. imo the localizations for a given feature should be added in the PR where the feature itself is added when possible.

vgskye:
fair, lemme add that after a rebase

gamma-delta:
If it's tested and confirmed working it LGTM.

gamma-delta (approved):
Yep, LGTM

vgskye:
why the fuck is hexdoc failing I didn't even touch anything doc-related

gamma-delta:
The hexdoc issue was fixed https://github.com/hexdoc-dev/hexdoc/commit/9e1025c13a1f552d34cd7885471edb34fb821e63 so hopefully that should pass now

gamma-delta:
We *should* probably document this at some point but again not something I care to block on.

PR #739, opened 2024-08-24
Check for Sprint bind instead of Ctrl for ShiftScrollListener

### LINKED ISSUE #179: [BUG] Abacus shift-scrolling doesn't work when sneak is bound to left control
Basically what the title says. Because "shift" scrolling uses the sneak keybind but "ctrl" scrolling is hardcoded to ctrl instead of using the run keybind, you can't set the abacus in increments of 1 or 0.1 if you set your sneak key to ctrl.

vgskye:
fixes #179.


object-Object:
how does this interact with the "toggle sprint" feature?

I wonder how much effort it would be to add a new keybind. Can we do that without conflicting with existing keys that are also bound to ctrl?

vgskye:
Toggle sprint seems to work the same as toggle sneak, turning ctrl into a toggle in here as well.

object-Object:
LGTM

PR #742, opened 2024-08-25
Document allay flay cost difference

### LINKED ISSUE #617: Need to update hexbook with regard to the now-recipe-dependent cost of Flay Mind
According to Talia, the cost of everything except the new Quenched Allay recipe is still 10 charged, but the text for Flay Mind makes it sound like it's 10 charged for everything, and all of the recipe pages (e.g. impeti) will still need to state their costs separately once the main page is fixed.

I can at least get the ball rolling on this tomorrow, though I can only reliably handle the US English translation (And of course that assumes my skill with my own first language improves relative what it was this morning. I weren't Englishing gud). Once I've submitted a PR, I'll leave it open for editing and we can go from there.

navarchus:
Fixes #617


navarchus:
I cannot for the life of me figure out how the image rendering in hexdoc works, so I cannot add an equivalent render to hexdoc like in game has after this PR.

gamma-delta:
I'd rather not block on Hexdoc of all things, if that's alright with everyone?

object-Object:
sure, I can do it in a separate PR if you want. I designed hexdoc to allow releasing web book updates separately from mod updates anyway.

gamma-delta:
if it doesn't work it's object's faultbject

gamma-delta:
@navarchus can you fix conflicts? then I'll merge it

object-Object:
added hexdoc

navarchus:
@gamma-delta done

Issue #640, opened 2024-03-26
Debug media cube is broken in pre-620


macks2008:
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-620

### Modloader version

Fabric: 0.1|.6 + API 0.91.0+ 1.20.1

### Modpack info

https://github.com/walksanatora/HexxyTestPack

### The latest.log file

https://mclo.gs/nvZYHdD

### Issue description

Despite my expectation, having a media cube renamed to "debug patterns" does not seem to do anything.

### Steps to reproduce

1. Load a creative world
2. spawn a media cube
3. rename the media cube "Debug Patterns" (any capitalization, last I checked, should work)
4. draw some patterns
5. nothing appears in chat, unlike on 1.19.2

### Other information

![2024-03-26_08 10 52](https://github.com/FallingColors/HexMod/assets/7110033/a6678ae7-5fb2-4f7c-8fb6-02220c8f791e)

macks2008:
the name I gave the media cube was enough for it to start doing the enchanted glow effect, so I must've done something right. It's just not working completely?

object-Object:
the exact name matters.

macks2008:
what is debug media supposed to do?  it didn't seem to do anything when I drew ol' Ray Caste
![image](https://github.com/FallingColors/HexMod/assets/7110033/42a3df55-4e33-4a76-b80a-04a5f3312abe)

object-Object:
make sure you're using it in survival - it's supposed to print all media usage drawn from the cube, so being in creative or having other sources messes with it

macks2008:
Okay it does work in survival. So it's just debug patterns that's not working, as far as I can tell.

object-Object:
great, thanks for confirming. I think it should be an easy fix then - hopefully just re-add the code that was removed.

macks2008:
🥳

gamma-delta:
I removed the code while I was refactoring the evaluator because I didn't actually know it was there (Wire added it without telling me) so I didn't know how it worked.

macks2008:
ah... whoops. That reminds me of a joke about code documentation, "when I wrote this code, only God and I knew what it did. Now only God knows". (Side note: what is Wire? Is that an IDE or gradle plugin or something?)
[...]

PR #743, opened 2024-08-26
Document Media Cube

### LINKED ISSUE #390: Hidden debug cube features should be documented somewhere
Currently, as far as I know, to find out about the debug cube features (eg. "debug media") you need to actually look at the code. I think there should be a table in the notebook listing them, maybe locked behind a hidden advancement granted by obtaining the media cube.

navarchus:
Fixes #390 
Add hidden entry in book gated by advancement. Explains function of Media Cube once it has been put into Player's inventory.


object-Object:
something seems messed up with the commits - there are some commits on this branch that are also on `main` but with a different SHA.

navarchus:
> something seems messed up with the commits - there are some commits on this branch that are also on `main` but with a different SHA.

I did a force push to revert an unnecessary commit, should I just open a new PR

object-Object:
you can if you want, or you could fix it with rebase, or I can probably fix it if I get a chance

navarchus:
> you can if you want, or you could fix it with rebase, or I can probably fix it if I get a chance

I'll give it a shot tomorrow

gamma-delta:
I feel like if we're documenting the media cube we should probably document the stuff it *does* (does it still have the thing where it displays all the patterns executed based on the anvil rename?)

object-Object:
`debug patterns` still works, but `debug media` doesn't (#640)

gamma-delta:
Again this is something I'd rather not block the release on. Hexdebug is a way better debugger anyways.

object-Object:
mmm, yeah. the point of #390 was specifically to document those hidden features accessible by renaming the item.

navarchus:
I can take a second pass at this and redo the documentation sometime this week if that's necessary.

Issue #249, opened 2022-09-17
[Suggestion] Expand Ignite Block's functionality


GameGuruGreenGryphon:
Ignite Block currently sets fire to the top of the vector passed into it
I suggest having Ignite Block take in more than one kind of input, and renamed to Ignite
This adds more things Hex is capable of doing while also saving patterns
This would probably require a media cost change, possibly a dynamic media cost depending on the action performed

Some ideas I've had...
Passing a vector into Ignite sets fire to the top of the block
Passing a vector which has a furnace instantly smelts an amount of items in the furnace
Passing an entity into Ignite sets fire to the entity directly, rather than needing to place fire beneath the entity
Passing an item entity into Ignite smelts the item if it can be smelted

PR #746, opened 2024-08-26
Make Ignite spell compatible with entities

miyucomics:
Fixes [this suggestion](https://github.com/FallingColors/HexMod/issues/249) by making the Ignite Block spell able to ignite entities for eight seconds ( the vanilla duration for entities walking into fire ). It renames Ignite Block to just Ignite, adds a new argument type for "an entity *or* a vector", and changes the description for the spell to describe its new capabilities.


Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
object-Object, line 1: Seems like there are a bunch of unnecessary whitespace changes in this file - can  you revert those?

object-Object (changes requested):
One comment about formatting, then I think LGTM.

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
miyucomics: I don't think I received an email for this so sorry for the late response. I think IntelliJ automatically formatted the file and I have reversed those changes.

object-Object (approved):

PR #747, opened 2024-08-27
Update phrasing on Cleric Impetus

### LINKED ISSUE #738: Unbound cleric circles are not documented
We should document that it's now possible to use a cleric impetus without a bound player.

This is the current text:

> A [Cleric Impetus](https://hexcasting.hexxy.media/v/0.11.1-7/1.0/en_us/#greatwork/impetus) must be bound to a player by using an item with a reference to that player, like a [Focus](https://hexcasting.hexxy.media/v/0.11.1-7/1.0/en_us/#items/focus), on the block.

navarchus:
Fixes #738


gamma-delta:
LGTM

PR #759, opened 2024-09-10
Remove reference to nonexistent Combination Distillation

### LINKED ISSUE #753: Disjunction Dstl entry still mentions Combination Dstl even though it no longer exists
### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-669

### Issue description

The book entry for Disjunction Distillation says the following:
> With two lists, this creates a list of every element from the first list, plus every element from the second list that is not in the first list. This is similar to [Combination Distillation](https://hexcasting.hexxy.media/v/0.11.1-7/1.0/en_us/#patterns/lists@hexcasting:add).

However, Combination Distillation no longer exists. The link does send you to the correct page (the Additive Distillation entry in the Lists chapter) but the text should probably be changed.

Robotgiggle:
Fixes #753 by tweaking the description of Disjunction Distillation to mention Additive Distillation rather than Combination Distillation. Also slightly rewords the sentence to make it more clear that Disjunction Distillation and Additive Distillation are not quite the same thing.


object-Object:
thanks!

PR #766, opened 2024-10-02
Document Media Cube/Creative Unlocker

### LINKED ISSUE #390: Hidden debug cube features should be documented somewhere

navarchus:
Fixes #390 
Add hidden entry in book gated by advancement. Explains function of Media Cube once it has been put into Player's inventory.


navarchus:
Re-created PR due to awful git commit history

PR #767, opened 2024-10-02
Extend ambit by imperceptible amount to account for potential machine precision errors.

garyyo:
Add 0.00000000001 to "ambit radius squared" calculations to account for potential machine precision errors when calculating the ambit at specific angles.

This fixes not being able to move a sentinel by exactly 16 blocks in any direction. Same fix is applied to the circle sentinel calculation and the player's ambit, and might be relevant to a bound cleric circle but I have not included that.

If you want to test the effect in the current codebase you can use the attached hex to do that (I comment the stack state a lot, if you want to remove that they are marked with a `///`).
[SentinelStepper.txt](https://github.com/user-attachments/files/17231248/SentinelStepper.txt)

It seems to be most likely to occur at shallow angles like pictured
image

### DIFF
--- a/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
+++ b/Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
@@ -1632,7 +1632,7 @@
         numlist: "Set operations are odd, in that some of them can accept two numbers or two lists, but not a combination thereof. Such arguments will be written as \"(num, num)|(list, list)\".$(br2)When numbers are used in those operations, they are being used as so-called binary \"bitsets\", lists of 1 and 0, true and false, \"on\" and \"off\".",
         
         "or.1": "Unifies two sets.",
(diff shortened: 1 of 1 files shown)


vgskye:
why not `Math.ulp(1.0)` (2.22e-16, also known as `double`'s Machine Epsilon), `net.minecraft.world.phys.shapes.Shapes.EPSILON` (1e-7), `com.mojang.math.Constants.EPSILON` (1e-6) or `net.minecraft.util.Mth.EPSILON` (1e-5)?

garyyo:
I did not know about them (expect ULP, which is too small). The (non-exhaustively) observed error is ~1e-12, so I chose an order of magnitude higher. If you think those are more appropriate I can use one of them, presumably the smallest one that still works is most appropriate (e.g. `net.minecraft.world.phys.shapes.Shapes.EPSILON`) but I don't know.

SamsTheNerd:
Machine epsilon is smallest precision in range [1,2) i think? then ulp just scales with the power of the double, so in the 16-32 squared range it'd be epsilon * (16^2) or so, which seems to still be a few orders of magnitude smaller than the value used here. I don't see any reason that this needs to be precise and we can't just overshoot it a tad, so i'm fine with using this value. 

I suppose the other concern would be very very large coordinate values giving an imprecise distance value? that seems harder to reason about though and so far seems to not be an issue, so i'm fine just going with this to solve the immediate issue of not being able to summon a sentinel 16 blocks away.

PR #771, opened 2024-10-26
Add pre-commit formatting

vgskye:


object-Object:
maybe we should also add a [pre-commit](https://pre-commit.com/) config? could also include ruff (Python formatting) for the hexdoc code

s5bug (changes requested):
Would be nice if there was an addition to the PR/CI action that caused failures if code was not formatted. I'm someone who 1. forgets to install pre-commit hooks 2. doesn't like pre-commit hooks and would rather run formatting manually anyway. I'm sure (1) will be very common for other people too.

PR #776, opened 2024-11-02
Make out of media a mishap and kill SideEffect earlyExit

### LINKED ISSUE #758: Circles behave strangely on out of media
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7.672

### Modloader version

_No response_

### Modpack info

_No response_

### The latest.log file

_No response_

### Issue description

As out of media is not a proper mishap but a SideEffect failure, it doesn't properly make the mishap sound, set a message, or abort the cast.

### Steps to reproduce

_No response_

### Other information

_No response_

vgskye:
supersedes #769.
Visible behaviour changes from this PR:
- Out of media correctly stops circle execution(closes #758)
- Out of media now turns the pattern red and chat messages are now tagged with the pattern's name
- Death from overcasting also shows the out of media message
- ~~Overcasting is more forgiving; Spells that would kill you and still fail no longer kill you~~
- ~~Other media overspending also no longer spends all your media~~

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
@@ -7,6 +7,7 @@ import at.petrak.hexcasting.api.casting.eval.vm.CastingImage
 import at.petrak.hexcasting.api.casting.eval.vm.SpellContinuation
 import at.petrak.hexcasting.api.casting.iota.Iota
 import at.petrak.hexcasting.api.casting.mishaps.MishapNotEnoughArgs
+import at.petrak.hexcasting.api.casting.mishaps.MishapNotEnoughMedia
 import at.petrak.hexcasting.common.lib.hex.HexEvalSounds
 
 /**
@@ -27,6 +28,8 @@ interface ConstMediaAction : Action {
     override fun operate(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation): OperationResult {
         val stack = image.stack.toMutableList()
 
+        if (env.extractMedia(this.mediaCost, true) > 0)
+            throw MishapNotEnoughMedia()
         if (this.argc > stack.size)
             throw MishapNotEnoughArgs(this.argc, stack.size)
         val args = stack.takeLast(this.argc)
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
@@ -29,7 +29,7 @@ interface ConstMediaAction : Action {
         val stack = image.stack.toMutableList()
 
         if (env.extractMedia(this.mediaCost, true) > 0)
-            throw MishapNotEnoughMedia()
+            throw MishapNotEnoughMedia(this.mediaCost)
         if (this.argc > stack.size)
             throw MishapNotEnoughArgs(this.argc, stack.size)
         val args = stack.takeLast(this.argc)
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
@@ -240,10 +240,10 @@ public abstract class CastingEnvironment {
      */
     public long extractMedia(long cost, boolean simulate) {
         for (var extractMediaComponent : preMediaExtract)
-            cost = extractMediaComponent.onExtractMedia(cost);
+            cost = extractMediaComponent.onExtractMedia(cost, simulate);
         cost = extractMediaEnvironment(cost, simulate);
         for (var extractMediaComponent : postMediaExtract)
-            cost = extractMediaComponent.onExtractMedia(cost);
+            cost = extractMediaComponent.onExtractMedia(cost, simulate);
         return cost;
     }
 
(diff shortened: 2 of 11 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -237,10 +237,10 @@ public boolean isEnlightened() {
      * If there was enough media found, it will return less or equal to zero; if there wasn't, it will be
      * positive.
      */
-    public long extractMedia(long cost) {
+    public long extractMedia(long cost, boolean simulate) {
         for (var extractMediaComponent : preMediaExtract)
```
SamsTheNerd, line 242: should the onExtractMedia events be called when simulated? perhaps there should be just a separate simulateExtractMedia?

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
```diff
@@ -27,6 +28,8 @@ interface ConstMediaAction : Action {
     override fun operate(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation): OperationResult {
         val stack = image.stack.toMutableList()
 
+        if (env.extractMedia(this.mediaCost, true) > 0)
```
SamsTheNerd, line 31: do we want to be so forgiving?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/sideeffects/OperatorSideEffect.kt
```diff
@@ -18,14 +18,11 @@ import net.minecraft.world.item.ItemStack
  */
 sealed class OperatorSideEffect {
     /** Return whether to cancel all further [OperatorSideEffect] */
-    abstract fun performEffect(harness: CastingVM): Boolean
+    abstract fun performEffect(harness: CastingVM)
```
SamsTheNerd, line 21: do we want to get rid of early exit entirely? I think I'm fine with errors being all mishap-based but it's possible some effect could want to early exit without loudly failing? I suppose that's less of an argument with the class being sealed. Seems like the sort of thing where if it was unsealed could maybe be useful for some addon somewhere maybe possibly?

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapNotEnoughMedia.kt
```diff
@@ -0,0 +1,21 @@
...
+    override fun accentColor(ctx: CastingEnvironment, errorCtx: Context): FrozenPigment =
+        dyeColor(DyeColor.RED)
+
+    override fun resolutionType(ctx: CastingEnvironment) = ResolvedPatternType.ERRORED
+
+    override fun execute(env: CastingEnvironment, errorCtx: Context, stack: MutableList<Iota>) {
+        env.extractMedia(cost, false)
+    }
+
+    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context) = "hexcasting.message.cant_overcast".asTranslatedComponent
```
SamsTheNerd, line 20: should this lang key be changed to be with the other mishap keys?

SamsTheNerd (changes requested):
otherwise looks fine to me

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
vgskye: I was thinking of having the extractMedia hooks also take a simulate argument

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
vgskye: huh? this precondition check in ConstMediaAction actually removes an exploit where you can use ConstMediaActions without enough media because media extraction failure happens after the action's actual execution iirc

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
SamsTheNerd: oh they do affect cost don't they. yeah that works for me

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/ConstMediaAction.kt
SamsTheNerd: oh that makes sense, i thought it was the overcasting more forgiving thing.

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/sideeffects/OperatorSideEffect.kt
vgskye: my interpretation of the SideEffect design is that preconditions should be checked before sideeffect operation, so earlyExit was a weird escape hatch out? I may be wrong here though

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapNotEnoughMedia.kt
vgskye: maybe. I didn't touch lang since it was not strictly required here

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
vgskye: done

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapNotEnoughMedia.kt
vgskye: probably out of scope for now

SamsTheNerd (commented):
code looks fine to me. PR should probably be tested in game (unclear if it was already or not). I'm not doing that rn, but poke me once that's done if I haven't gotten to it myself.

SamsTheNerd (approved):

PR #777, opened 2024-11-02
Make playerless sentinels mishap

### LINKED ISSUE #774: Sentinel patterns behave strangely on unbound cleric circles
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-679

### Issue description

Unlike Mind's Reflection, the various sentinel-related patterns do not have a check to see whether the casting entity exists. This results in some strange behavior when an unbound cleric circle attempts to cast them.

When the circle tries to cast Summon Sentinel or Summon Greater Sentinel, the slate produces mishap particles but does not actually trigger a mishap, and (presumably) fails to create a sentinel. Even worse, when it tries to caste Locate Sentinel, it _does_ mishap, and lists a java exception in the mishap message.

Attempting to cast Summon Sentinel

Attempting to cast Locate Sentinel

### Steps to reproduce

1. Create circle using unbound cleric impetus
2. Attempt to run `waystone rfln, summon sentinel, locate sentinel`
3. Observe that the Summon Sentinel slate produces mishap particles but does not actually mishap or stop the media wave
4. Observe that the Locate Sentinel slate does mishap, and that the cleric impetus lists a java exception as the cause

vgskye:
fixes #774. should the bad caster mishap be documented?

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadCaster.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadCaster.kt
@@ -1,0 +1,17 @@
+package at.petrak.hexcasting.api.casting.mishaps
+
+import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
+import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.pigment.FrozenPigment
+import net.minecraft.world.item.DyeColor
+
+class MishapBadCaster: Mishap() {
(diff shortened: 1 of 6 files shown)


SamsTheNerd (commented):
what happens in the case of non-player entities?

SamsTheNerd (changes requested):
should check that the casting entity is a player, since sentinel api requires a player. can do `if(env.castingEntity !is Player)` i believe?

SamsTheNerd:
otherwise looks good to me. Wouldn't hurt to document it, I'm not sure how consistent we are with that currently.

SamsTheNerd (approved):
made change

PR #779, opened 2024-11-03
Make Boolean Directrix Fisherman Directrix and document it

vgskye:
identical to #778 except the profession is fisherman, because shepherd feels a bit too generic and also is harder to level


SamsTheNerd:
went with shepherd

PR #787, opened 2024-11-12
Implement grid modifier potions

Robotgiggle:
Fully implements the previously unobtainable grid modifier potions, including recipes, proper effect icons, and a book entry.

image
image


Robotgiggle:
Moved to #788

PR #796, opened 2024-11-25
Blocky Chroniclers

SamsTheNerd:
Allows chroniclers to read/write blocks. 

Blocks can opt in in three ways.
1) implement `ADIotaProvider` on a Block Entity class.
2) implement `IBlockyIotaProvider` on a Block class.
3) manually register with HexAPI#registerBlockyIotaHolder

Option 1 will likely work best in most cases (since data is generally stored in a block entity), but option 2 is good for static reading or for reading that only depends on the environment (like akashic record). Finally, option 3 is provided for cases where you can't implement an interface, such as for another mod's block or for optional hex interop.


Common/src/main/java/at/petrak/hexcasting/api/addldata/ADIotaHolder.java
```diff
@@ -34,4 +34,24 @@ default Iota emptyIota() {
      * @return whether it is possible to write to this IotaHolder
      */
     boolean writeable();
+
+    // a helper for read-only holders
+    interface ADIotaHolderReadOnly extends ADIotaHolder{
```
object-Object, line 39: should there also be a WriteOnly version?

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBad.kt
```diff
@@ -0,0 +1,24 @@
+package at.petrak.hexcasting.api.casting.mishaps
+
+import com.mojang.datafixers.util.Either
+import net.minecraft.core.BlockPos
+import net.minecraft.world.entity.Entity
+import net.minecraft.world.item.ItemStack
+import net.minecraft.world.phys.Vec3
+
+// a silly stupid wrapper for throwing arbitrary badness mishaps (the ones that have a .of(thing, stub) constructor.
+class MishapBad {
```
object-Object, line 10: IMO this should just be a function in ActionUtils or something. Or if it has to be its own class for some reason, it should at least be abstract.

Common/src/main/java/at/petrak/hexcasting/common/blocks/akashic/BlockEntityAkashicBookshelf.java
```diff
@@ -89,4 +92,20 @@ protected void loadModData(CompoundTag tag) {
...
+        return IotaType.serialize(new PatternIota(pattern));
+    }
+
+    @Override
+    public Iota readIota(ServerLevel world) {
+        if(pattern == null){
+            return emptyIota();
+        }
+        return new PatternIota(pattern);
+    }
```
object-Object, line 110: It feels weird for the logic here to be duplicated... not sure how to fix it though.

Common/src/main/java/at/petrak/hexcasting/xplat/IXplatAbstractions.java
```diff
@@ -124,6 +125,11 @@ public interface IXplatAbstractions {
     @Nullable
     ADIotaHolder findDataHolder(Entity entity);
 
+    @Nullable
+    default ADIotaHolder findDataHolder(BlockPos pos, ServerLevel world){
+        return HexAPIImpl.getBlockyIotaHolder(world, pos);
+    }
+
```
object-Object, line 132: It feels kinda weird to put something in IXplatAbstractions if it's not actually different per-platform. I see why it's there, I guess... maybe add a comment justifying it?

Common/src/main/resources/assets/hexcasting/lang/ru_ru.flatten.json5
```diff
@@ -977,6 +977,12 @@
         "": "Ожидал %s при %s, но получил %s",
         sapling: "саженец",
         replaceable: "куда-нибудь поместить блок",
+        iota: {
+          "": "место для хранения йоты",
+          read: "место, откуда можно читать йоты",
+          write: "место, куда можно записывать йоты",
+          readonly: "место, которое будет принимать %s",
+        }
```
object-Object, line 985: where's this from?

Common/src/main/resources/assets/hexcasting/lang/zh_cn.flatten.json5
```diff
@@ -992,6 +992,12 @@
         "": "本应在%2$s处接受%1$s，而实际接受了%3$s",
         sapling: "一个树苗",
         replaceable: "一个可放置方块的地方",
+        iota: {
+          "": "一个可以存储iota的地方",
+          read: "一个可以读出iota的地方",
+          write: "一个可以写入iota的地方",
+          readonly: "一个能够接受%s的地方",
+        }
```
object-Object, line 1000: where's this from?

Common/src/main/java/at/petrak/hexcasting/common/blocks/circles/BlockEntitySlate.java
```diff
@@ -42,4 +48,39 @@ protected void loadModData(CompoundTag tag) {
...
+            if(iota instanceof PatternIota pIota){
+                this.pattern = pIota.getPattern();
+                sync();
+            }
+            if(iota instanceof NullIota || iota == null){
+                this.pattern = null;
+                sync();
+            }
+        }
+        return iota instanceof PatternIota || iota instanceof NullIota || iota == null;
```
object-Object, line 79: This feels weirdly duplicated, but idk how to fix it (~~other than moving to Kotlin~~).

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpIgnite.kt
```diff
@@ -30,27 +27,22 @@ object OpIgnite : SpellAction {
...
-            }
-            else -> throw MishapInvalidIota.ofType(target, 0, "entity_or_vector")
-        }
+        val target = args.getEntityOrBlockPos(0, argc)
+
+        target.map(env::assertEntityInRange, env::assertPosInRange)
+
+        val burstPos = target.map({ent -> if (ent is ItemEntity) {
+            // Special case these because the render is way above the entity
+            ent.position().add(0.0, 3.0 / 8.0, 0.0)
```
object-Object, line 36: maybe this logic could be moved to a utility function? eg `ParticleSpray.getEntityBurstPos` or something

object-Object (changes requested):
Some nitpicks, some questions. Otherwise I think it overall looks good, assuming you've tested everything.

object-Object:
Moved to https://github.com/object-Object/IoticBlocks since Petra doesn't want it in the base mod.

PR #806, opened 2024-11-28
Fix continuation-related exploits and bump* stack limit

vgskye:
Bumps stack size limit to 8k, but includes extra iotas stored in current CastingImage in the tally(escaped iotas and continuations).
Also fixes iris truename smuggling by properly implementing subIotas.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ContinuationIota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ContinuationIota.java
@@ -13,6 +13,7 @@ import net.minecraft.nbt.Tag;
 import net.minecraft.network.chat.Component;
 import net.minecraft.server.level.ServerLevel;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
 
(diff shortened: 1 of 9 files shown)


gamma-delta:
I kind of like using Iris to smuggle truenames as an *idea* because it's so clever, but I'm not sure if it's bad or good for gameplay ... what do you all think?
Will also ask on the discord.

vgskye:
converting to draft because this needs performance regression testing

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ContinuationIota.java
```diff
@@ -56,17 +57,19 @@ public boolean executable() {
...
     @Override
     public int size() {
-        var continuation = this.getContinuation();
-        var size = 0;
-        while (continuation instanceof SpellContinuation.NotDone notDone) {
-            size += 1;
-            size += notDone.component1().size();
-            continuation = notDone.component2();
-        }
+        return Math.min(this.getContinuation().size(), 1);
```
object-Object, line 67: this should be Math.max, as discussed on Discord - min results in the value only being 0 or 1, max results in the value being >=1.

Issue #783, opened 2024-11-08
Place Block requires a block in the hotbar rather than pulling from Inventory


garyyo:
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-685

### Modloader version

fabric-1.20.1

### Issue description

Similar to #671, seems that placeblock does not take from your inventory as described in the book when the block to use is in your hotbar, and mishaps (tossing your casting items out of your hands) when you do have enough blocks in your inventory, but not enough in your hotbar (which the video showcases). 

Likewise does not seem to work correct if there is a single block is in your offhand and the rest in your inventory (At first I thought it did but I switched to creative mode to test which gave me the wrong results). 

https://github.com/user-attachments/assets/812eda08-6a66-4249-874a-1944511c1686

### Steps to reproduce

1. Place 1 block in your hotbar (or offhand), and more in the rest of your inventory.
2. Do place block twice in different locations in the same hex
3. Cast that hex

PR #812, opened 2024-12-02
Fix placeblock not pulling items from inventory when they are available.

### LINKED ISSUE #671: Place Block requires a block in the offhand rather than pulling from inv/hotbar
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.1-7-pre-634

### Modloader version

Fabric: loader 0.15.9, api 0.92.1+1.20.1

### Modpack info

HexxyTest (https://github.com/walksanatora/HexxyTestPack)

### Issue description

In the past, Place Block would select a block to place based on the options available in your hotbar, as described in the Working With Items section of the book. Now, however, Place Block only seems to be able to place from your offhand. Attempting to cast it without a placeable block in your offhand fails, along with providing a badly formatted mishap message. This makes complex block placement essentially impossible without some way to change what's in your offhand mid-cast.

### Steps to reproduce

1. Attempt to cast Place Block without a placeable item in your offhand
2. Observe that the spell fails, and produces the following mishap message
image

### Other information

_No response_

garyyo:
Fixes #671, whoops actually #783
The predicate now tests if objects in the inventory are the same item same tags instead of same object ref.


gamma-delta:
:facepalm: ah, kotlin equals semantics

PR #823, opened 2025-01-10
Add player attributes from Hexxy Attributes

beholderface:
This code is largely based on that of Hexxy Attributes by justs_js, PR'd with his permission.
https://discord.com/channels/936370934292549712/1011455473528098857/1327275691343085578

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/common/items/ItemStaff.java
+++ b/Common/src/main/java/at/petrak/hexcasting/common/items/ItemStaff.java
@@ -1,6 +1,7 @@
 package at.petrak.hexcasting.common.items;
 
 import at.petrak.hexcasting.api.HexAPI;
+import at.petrak.hexcasting.common.lib.HexAttributes;
 import at.petrak.hexcasting.common.lib.HexSounds;
 import at.petrak.hexcasting.common.msgs.MsgClearSpiralPatternsS2C;
 import at.petrak.hexcasting.common.msgs.MsgOpenSpellGuiS2C;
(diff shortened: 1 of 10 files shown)


gamma-delta:
Can I get a second pair of eyes on this? @object-Object @SamsTheNerd ?

Common/src/main/java/at/petrak/hexcasting/common/items/ItemStaff.java
```diff
@@ -25,6 +26,9 @@ public ItemStaff(Properties pProperties) {
 
     @Override
     public InteractionResultHolder<ItemStack> use(Level world, Player player, InteractionHand hand) {
+        if (player.getAttributeValue(HexAttributes.FEEBLE_MIND) > 0){
```
SamsTheNerd, line 29: Is this fine on client?

Fabric/src/main/java/at/petrak/hexcasting/fabric/mixin/FabricPlayerMixin.java
```diff
@@ -22,5 +22,9 @@ protected FabricPlayerMixin(EntityType<? extends LivingEntity> entityType, Level
         var out = cir.getReturnValue();
         out.add(HexAttributes.GRID_ZOOM);
         out.add(HexAttributes.SCRY_SIGHT);
+        out.add(HexAttributes.FEEBLE_MIND);
+        out.add(HexAttributes.MEDIA_CONSUMPTION_MODIFIER);
+        out.add(HexAttributes.AMBIT_RADIUS);
+        out.add(HexAttributes.SENTINEL_RADIUS);
```
SamsTheNerd, line 28: Should these be like,, auto grabbed out of the HexAttributes instead of manually listed in 3 separate places?

Common/src/main/java/at/petrak/hexcasting/common/items/ItemStaff.java
SamsTheNerd: Nvm, I think it should be fine since they're marked syncable?

SamsTheNerd (approved):
Have like one tiny maybe nitpick that's probably fine to ignore but otherwise it looks fine to me

PR #818, opened 2024-12-13
Loot cyphers, scroll tweaks, and more

Robotgiggle:
- Adds ancient cyphers, which can be found in a variety of loot tables and come preloaded with a random hex
  - Ancient cyphers look like normal cyphers but with tarnished copper and slightly faded amethyst, and work identically
  - Unlike normal cyphers, they display their stored hex in the tooltip, so they can serve as a tutorial for new players
  - You can also erase them and put in your own spells using Craft Cypher
- Adds various new things related to Ancient Scrolls
  - You can now set the `op_id` without setting the `pattern`, which creates a scroll that loads your world's version of the pattern when ticked
  - There's now a creative mode tab containing all of the scrolls, including ones from addons, using the above feature to load the correct pattern when you grab one and put it in your inventory
  - The wandering trader now has a chance to sell one random scroll for 12 emeralds, which only reveals its stroke order when you buy it (finally, a way to get scrolls in skyblock!)
- Adds two custom loot tables (`hexcasting:random_scroll` and `hexcasting:random_cypher`) that produce a guaranteed scroll or ancient cypher when generated
  - These aren't used anywhere in game, they're intended for use by packmakers or addon devs
- Assorted new config options
  - Injection list and injection chance for ancient cyphers, just like how it works for lore fragments
  - List of preset hexes for ancient cyphers
  - Chance for wandering trader to sell a scroll
  - Toggle to enable comma-separated pattern lists

PR #839, opened 2025-01-23
Misc changes

Robotgiggle:
All the other random changes from #818, and some extra stuff
- Two custom loot tables (`hexcasting:random_scroll` and `hexcasting:random_cypher`) that produce a guaranteed scroll or ancient cypher when generated, for use by packmakers and addon devs
- Config toggle to enable comma-separated pattern lists
- The book now gives the correct cost for Wayfarer's Flight (1 shard/s, not 1 charged/s)
- Slightly reworded the description of Ignite to clarify that it works on all entities
- Minor cleanup in the forge config file


object-Object:
Merging via #844 instead.

PR #841, opened 2025-01-29
Config option for Greater Teleport item splatting

### LINKED ISSUE #502: [Feature request] Make greater teleport dropping items into a gamerule and/or config option
Some modpacks contain mods with other means of long-range teleportation. In case of the modpack creator desiring to limit such features, usually it can be achieved by limiting crafting recipes or config. 
In the opposite case of modpack creator desiring to enable such features, HexCasting suddenly falls behind due to the crippling downside of making a mess of dropped items not limited by any server configuration. This becomes even worse when multiple people are using the teleport either at the same time or in a short span of time.

TheDrawingCoder-Gamer:
adds config option for whether items fly out of your inventory upon using greater teleport

closes #502

### DIFF
--- a/Forge/src/main/java/at/petrak/hexcasting/forge/ForgeHexConfig.java
+++ b/Forge/src/main/java/at/petrak/hexcasting/forge/ForgeHexConfig.java
@@ -137,6 +137,8 @@ public class ForgeHexConfig implements HexConfig.CommonConfigAccess {
         private static ForgeConfigSpec.IntValue maxSpellCircleLength;
         private static ForgeConfigSpec.ConfigValue<List<? extends String>> actionDenyList;
         private static ForgeConfigSpec.ConfigValue<List<? extends String>> circleActionDenyList;
+
+        private static ForgeConfigSpec.BooleanValue greaterTeleportSplatsItems;
         private static ForgeConfigSpec.BooleanValue villagersOffendedByMindMurder;
         private static ForgeConfigSpec.ConfigValue<List<? extends String>> tpDimDenyList;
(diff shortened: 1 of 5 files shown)


Forge/src/main/java/at/petrak/hexcasting/forge/ForgeHexConfig.java
```diff
@@ -175,6 +176,10 @@ public Server(ForgeConfigSpec.Builder builder) {
                     "Resource locations of disallowed actions. Trying to cast one of these will result in a mishap.")
                 .defineList("actionDenyList", List.of(), Server::isValidReslocArg);
 
+            greaterTeleportSplatsItems = builder.comment(
+                    "Should greater teleport splat players inventory?"
```
object-Object:
```suggestion
                    "Should items fly out of the player's inventory when using Greater Teleport?"
```

imo since there's no title here, it's better to say exactly what it does (copied from Fabric) than to keep the whimsical "splat"

object-Object (approved):

object-Object:
oh, it seems like there are some conflicts now

TheDrawingCoder-Gamer:
do I rebase and close and reopen this PR?

object-Object:
a merge commit would have been better, but I guess that works

object-Object:
you shouldn't need to close and reopen the PR.

TheDrawingCoder-Gamer:
Oops, broke the commit I just applied in github, LOL

object-Object:
I haven't tested ingame after the rebase, but I'm going to lazily merge it in anyway because I don't see why it would have broken (fingers crossed). Thanks for the PR!

PR #843, opened 2025-02-03
Remove The Media Cube category; move The Media Cube entry into Items and make it secret

object-Object:
This PR removes the entirely unnecessary dedicated category for the media cube, instead moving it into the Items category. It also sets the media cube entry to secret, so it doesn't show up as a locked entry in the book for players who have not picked up a media cube.


SamsTheNerd (commented):
should it have a showcase page? i don't think patchi does the little hold ctrl to open page thing unless you have a showcase or a recipe page and it doesn't have a recipe. Otherwise looks fine to me.

object-Object:
I couldn't test the ctrl thing, it didn't seem to work on any Hex items in the dev env.

PR #850, opened 2025-02-08
Stop fake players from bloodcasting

beholderface:
People on the discord have recently discovered that Create deployers in 1.20 can gain access to bloodcasting. Since their fakeplayer entities don't track health like real players do, this leads to the deployer essentially getting two free charged worth of media with every single cast. I do not think this exploit should be allowed to persist.


beholderface:
👍

PR #860, opened 2025-02-21
Update item rarities

Robotgiggle:
Adds [rarity](https://minecraft.wiki/w/Rarity) to a few more items. Specifically:
- Filled impeti and directrices, along with the mindsplice staff, are now uncommon
- Quenched shards, along with the quenched staff, are now uncommon
- Quenched blocks are now rare

Also changes the rarity of the Amethyst Sconce back to common - why on earth was an easy-to-craft decorative block marked as rare?

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/common/lib/HexBlocks.java
+++ b/Common/src/main/java/at/petrak/hexcasting/common/lib/HexBlocks.java
@@ -163,15 +163,18 @@ public class HexBlocks {
     public static final BlockRightClickImpetus IMPETUS_RIGHTCLICK = blockItem("impetus/rightclick",
         new BlockRightClickImpetus(slateish()
             .pushReaction(PushReaction.BLOCK)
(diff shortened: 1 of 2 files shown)


Common/src/main/java/at/petrak/hexcasting/common/lib/HexBlocks.java
```diff
@@ -192,7 +197,10 @@ private static BlockBehaviour.Properties quenched() {
     public static final BlockAkashicLigature AKASHIC_LIGATURE = blockItem("akashic_connector",
         new BlockAkashicLigature(akashicWoodyHard().lightLevel(bs -> 4)));
 
-    public static final BlockQuenchedAllay QUENCHED_ALLAY = blockItem("quenched_allay", new BlockQuenchedAllay(quenched()));
+    public static final BlockQuenchedAllay QUENCHED_ALLAY = blockItem("quenched_allay", 
+        new BlockQuenchedAllay(quenched()), 
+        HexItems.props().rarity(Rarity.RARE)
```
Alwinfy: just uncommon perhaps

PR #861, opened 2025-02-21
Document bool directrix mishap

Robotgiggle:
Adds a page in the Enlightened Mishaps section describing what happens if you provide something that isn't a bool (or nothing at all) to a Shepherd Directrix. Internally these are two separate mishaps, but the effect is exactly the same so I think this single page covers both cases adequately.

image


Robotgiggle:
Moved to #862

PR #866, opened 2025-02-24
Optimizes Spell Circles

Stick404:
In 1.20 Spell Circles can loop on their own slate, however this can cause massive NBT-Leaks due to the same positions stacking on the same positions. This PR just makes reachedPositions a HashSet/Set, and adds an extra Int Var of "reachedNumber", or how many slates it has come across (this is for the speed up of Spell Circles).

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
@@ -4,7 +4,9 @@ import at.petrak.hexcasting.api.HexAPI;
 import at.petrak.hexcasting.api.casting.eval.env.CircleCastEnv;
 import at.petrak.hexcasting.api.casting.eval.vm.CastingImage;
 import at.petrak.hexcasting.api.misc.Result;
+import at.petrak.hexcasting.api.mod.HexConfig;
 import at.petrak.hexcasting.api.pigment.FrozenPigment;
+import at.petrak.hexcasting.api.utils.ChunkScanning;
 import at.petrak.hexcasting.api.utils.HexUtils;
 import com.mojang.datafixers.util.Pair;
 import net.minecraft.ChatFormatting;
@@ -15,8 +17,7 @@ import net.minecraft.nbt.ListTag;
 import net.minecraft.nbt.NbtUtils;
 import net.minecraft.nbt.Tag;
 import net.minecraft.network.chat.Component;
-import net.minecraft.server.level.ServerLevel;
-import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.server.level.*;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
 import net.minecraft.world.phys.AABB;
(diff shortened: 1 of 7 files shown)


Stick404:
A total list of things changed:
* Removes the memory/NBT leak of Spell Circles
* Stores only the highest and lowest vectors of the Spell Circle bounds, rather than all found possible slate
* Removes an old Null Logger in Cleric Impetus
* Optimized Spell Circle scanning, instead of calling the normal getBlockState, which loads many more chunks, it now calls getChunkFuture. In case the future fails/does not work, it still calls getBlockState as a fallback
* Now respects the Slate Limit in the config

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -31,42 +36,53 @@ public class CircleExecutionState {
...
-    public final Set<BlockPos> knownPositions;
-    public final List<BlockPos> reachedPositions;
+    public final HashSet<BlockPos> reachedPositions;
     public BlockPos currentPos;
     public Direction enteredFrom;
     public CastingImage currentImage;
     public @Nullable UUID caster;
     public @Nullable FrozenPigment casterPigment;
+    // This controls the speed of the current slate
+    public Long reachedSlate;
```
vgskye: any reason this is Long and not long?

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -94,14 +110,36 @@ protected CircleExecutionState(BlockPos impetusPos, Direction impetusDir, Set<Bl
...
         var seenGoodPosSet = new HashSet<BlockPos>();
-        var seenGoodPositions = new ArrayList<BlockPos>();
-
+        var positiveBlock = new BlockPos.MutableBlockPos();
+        var negativeBlock = new BlockPos.MutableBlockPos();
+        var lastBlockPos = new BlockPos.MutableBlockPos();
+        BlockPos firstBlock = null;
+        HashMap<ChunkPos, ChunkAccess> chunkMap = new HashMap<>();
         while (!todo.isEmpty()) {
+
```
vgskye, line 114: superfluous formatting change

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -231,6 +294,7 @@ public boolean tick(BlockEntityAbstractImpetus impetus) {
 
         executorBlockState = executor.startEnergized(this.currentPos, executorBlockState, world);
         this.reachedPositions.add(this.currentPos);
+        this.reachedSlate = this.reachedSlate +1;
```
vgskye: `this.reachedSlate += 1;`

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -257,7 +321,7 @@ public boolean tick(BlockEntityAbstractImpetus impetus) {
                                 Component.literal(this.currentPos.toShortString()).withStyle(ChatFormatting.RED)),
                             new ItemStack(Items.COMPASS));
                         ICircleComponent.sfx(this.currentPos, executorBlockState, world,
-                            Objects.requireNonNull(env.getImpetus()), false);
+                                Objects.requireNonNull(env.getImpetus()), false);
```
vgskye: superfluous indent change

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -289,7 +353,7 @@ public boolean tick(BlockEntityAbstractImpetus impetus) {
      * How many ticks should pass between activations, given the number of blocks encountered so far.
      */
     protected int getTickSpeed() {
-        return Math.max(2, 10 - (this.reachedPositions.size() - 1) / 3);
+        return Math.toIntExact(Math.max(2, 10 - (this.reachedSlate - 1) / 3));
```
vgskye: didn't even know `Math.toIntExact` was a thing. any reason for doing this instead of a normal cast-to-int?

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
Stick404: Misinput, fixed

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
Stick404: Strange recommendation from IDEA, fixed

Stick404:
Went through these and fixed the local version. I'll make a push once there are no more random problems with the code

Stick404:
This has been split into 3 PRs
#908 
#909 
#910

PR #879, opened 2025-04-02
pattern overlays

SamsTheNerd:
image
it does this when you shift. tested on forge and fabric, most of the work is done by inline's render hook. updated inline to be able to use it.


Common/src/main/java/at/petrak/hexcasting/interop/inline/HexPatternOverlayRenderer.java
```diff
@@ -0,0 +1,55 @@
...
+import at.petrak.hexcasting.common.items.storage.ItemScroll;
+import at.petrak.hexcasting.common.items.storage.ItemSlate;
+import com.samsthenerd.inline.api.client.extrahooks.ItemOverlayRenderer;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.client.gui.screens.Screen;
+import net.minecraft.world.item.ItemStack;
+
+import java.util.function.Function;
+
+public record HexPatternOverlayRenderer(Function<ItemStack, HexPattern> patternRetriever) implements ItemOverlayRenderer {
```
Alwinfy: nit: why is this a record holding a FI and not just a FI itself? i.e.

```
@FunctionalInterface
public interface HexPatternOverlayRenderer extends ItemOverlayRenderer {
  HexPattern getPattern(ItemStack stack);
}
```

SamsTheNerd: Don't remember, prob just didn't think of it. Want me to change it ?

Robotgiggle (approved):
Made the change Alwinfy suggested - LGTM

PR #885, opened 2025-04-26
Make akashic ligatures able to use connected textures.

kineticneticat:
I got bored and did this mainly because I have wanted it for ages, and I think it looks neat.
The textures are a bit repeat-y, but I think it looks fine, though I could always have another go at it

image

If this isn't wanted, it's fine (again, I was just bored), and I can just stick it in a resource pack for myself, I think.
Also, since it uses continuity, it's a soft dependency and completely client-side (and should also work with optifine as they use the same format, I think).

P.S. I did have to bump the versions of fabric/fabric api/etc to make continuity 3.0.0 compatible, but I'm not sure why they aren't bumped by default, as hex seemed to work fine with them (I didn't include this change as I assumed there was a reason for it).


Robotgiggle (approved):
LGTM

PR #895, opened 2025-06-18
HexCasting port to 1.21 (NeoForge part)

slava110:
Fabric is commented out.
This PR is based on my NeoForge repo of HexCasting. Some NeoForge leaked to XPlat. This should be easily fixable though.
# Requirements
This port requires version of Paucal with https://github.com/gamma-delta/PAUCAL/pull/12 and https://github.com/gamma-delta/PAUCAL/pull/13. You can build it yourself, publish to maven (local or remote).
# Notable changes
- Added Codecs to everything c:
- Tag representation of iotas has been removed. They are deserialized using Codecs, accessed in ItemStacks through data components, in block entity via field.
- EntityIota is reference-based now. Instead of directly referencing Entity object, its stores entity UUID and cached actual entity. This mechanism allows safely unloading and loading entities again without losing reference to them. Besides that it's just simply a necessity with Codecs.
- All `List<Iota>.getEntity` methods now require `ServerLevel` as argument due to the change above.
- Garbage and Null iotas are actually singletons now
- Iotas now store values inside subtypes. Hashcode is marked as abstract in superclass so users know that it should be overriden.
# Requires attention
- Rendering-related anything. I'm not an expert in rendering and `VCDrawHelper` might have been ported poorly
- Datacomponent-related anything.


Sirius902:
Seems like Scribe's Reflection no longer works on Ancient Scrolls. I gave myself one with `/hexcasting perWorldPatterns give hexcasting:lightning` (which interestingly also did seem to crash the game with `Internal Exception: io.netty.handler.codec.EncoderException: Failed to encode packet 'clientbound/minecraft:system_chat'`).
image

object-Object:
Closing in favour of #985

PR #900, opened 2025-07-14
Added Harmony Reflection

miyucomics:
Adds a reflection for the golden ratio. It seems fitting because there is a pattern for pi and a separate one for tau, so why not one for phi.


Common/src/main/java/at/petrak/hexcasting/common/lib/hex/HexActions.java
```diff
@@ -464,6 +464,9 @@ public class HexActions {
     public static final ActionRegistryEntry CONST$E = make("const/double/e",
         new ActionRegistryEntry(HexPattern.fromAngles("aaq",
             HexDir.EAST), Action.makeConstantOp(new DoubleIota(Math.E))));
+    public static final ActionRegistryEntry CONST$PHI = make("const/double/phi",
+        new ActionRegistryEntry(HexPattern.fromAngles("wdded",
+            HexDir.NORTH_EAST), Action.makeConstantOp(new DoubleIota(1.618))));
```
vgskye: `1.618033988749895` :3

miyucomics:
There are (virtually) infinite pattern slots, right? I know phi is easy to compute, but so is e via its Taylor series and that made it into the standard library. Tau is there, I assume because it's the more "pure" circle constant, and pi is there, I assume because people don't want to have to do `tau 2 div` each time. By that logic, why not phi too? It's mechanically trivial to compute, yes, but noisy in context when you look over a hex and have to lexiconize `5 0.5 pow 1 add 0.5 mul` as "the patterns that push the golden ratio." It saves players the operations it would take to calculate that number by just having it as a reflection.

I'd actually say that e has fewer gameplay applications than phi for most people. Phi is nice for geometry. I wanted to add it because I was working on a little project (scattering points evenly across a sphere using the Fibonacci sphere technique in order to send raycasts off in all directions) and the golden ratio as a readily-accessible pattern would have been really helpful. There is also the more famous use of sunflower-like configurations, spiral tilings, and phyllotaxis. Even if phi isn't frqeuently used, it's just a nice number that encapsulates a nice idea.

Sure, one could work around it by embedding or computing it but that is so much overhead for a constant that could just be baked in. The pull request is already here, it isn't competing with any more useful patterns, it saves op counts like Dioscuri's Gambit, and it might nudge players towards some fun uses. There isn't any harm I see in including it and the upside is small and sweet.

Common/src/main/java/at/petrak/hexcasting/common/lib/hex/HexActions.java
miyucomics: I only added three decimal places because of Hex's double tolerance, but I realize that you could probably scale it up to get more digits. I'll update it.

miyucomics:
Question about PRing: should I be running datagen and pushing the altered datagen files?

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -1788,6 +1789,7 @@
           tau: "Adds τ, the radial representation of a complete circle, to the stack.",
           pi: "Adds π, the radial representation of half a circle, to the stack.",
           "e": "Adds $(italic)e/$, the base of natural logarithms, to the stack.",
+          "phi": "Adds $(italic)φ/$, the ratio of a whole to its greater part, to the stack.",
```
PoolloverNathan: * description probably needs some work (tentative description suggested; what does the golden ratio do again?)
* I'd unquote phi (and e) for consistency
* other greek letters aren't italicized
```suggestion
          e: "Adds $(italic)e/$, the base of natural logarithms, to the stack.",
          phi: "Adds φ, the golden ratio, to the stack.",
```

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -695,6 +695,7 @@
             pi: "Arc's Reflection",
             tau: "Circle's Reflection",
             "e": "Euler's Reflection",
+            "phi": "Harmony Reflection",
```
PoolloverNathan:
```suggestion
            e: "Euler's Reflection",
            phi: "Harmony Reflection",
```

PoolloverNathan (changes requested):
looks good, but some minor formatting nitpicks (and the description)

Robotgiggle:
I made most of pool's suggested lang tweaks, but `"e"` remains quoted because otherwise JSON thinks it's the scientific notation operator.

Twisted-Code:
FWIW (as somebody late to the party but still adding my two cents anyway) I feel like baking this kind of thing in is generally the right call. Hex Casting as an esoteric language remains esoteric without requiring people to repeatedly compute well-known constants. There are limits, of course, but I do not think this is *necessarily* crossing the line. 
That said, if you really want to keep it a puzzle even for well-established things like this, the only thing I might suggest is locking constant patterns like this behind an advancement that checks (perhaps once every 20 ticks) whether a number that's "close enough" (within epsilon of the constant) is on the top of the stack. 
As for a scribe's Easter egg: Scribe's reflection on anything tagged with `minecraft:piglin_loved`? Or does that include non-golden items? (Even if so, semantically it makes sense to me, taking golden to mean beautiful)

PR #902, opened 2025-07-14
fix: the busted bulleting in README.md

Alwinfy:
shits been bothering me for years


object-Object:
can you also remove the duplicate curseforge link while you're at it? :p

PR #907, opened 2025-07-14
Make public some casting environment methods

miyucomics:
I can think of at least a few cases in addons where a person would want to perform their own search of what a casting environment has to offer. Features in base generally shouldn't be done just for addons but I don't see why these very useful methods should be protected and it's currently impossible to access them with usual techniques like invokers because one of the parameters is also a protected enum. Access wideners also don't work because those exclusively only work for base Minecraft classes, not other mods.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
@@ -376,7 +376,7 @@ public abstract class CastingEnvironment {
     /**
      * Get all the item stacks this env can use.
      */
-    protected abstract List<ItemStack> getUsableStacks(StackDiscoveryMode mode);
+    public abstract List<ItemStack> getUsableStacks(StackDiscoveryMode mode);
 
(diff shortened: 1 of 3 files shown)


vgskye (approved):
I mean, it's probably possible to use ugly reflection to use these methods, so it's not completely impossible, but yeah LGTM

object-Object:
hmm, should we count this as a breaking change? it seems like it [won't affect existing mod builds](https://docs.oracle.com/javase/specs/jls/se17/html/jls-13.html#jls-13.4.7), but I think any addon that implements these methods would get a compilation error after this change.

poolcritter:
> hmm, should we count this as a breaking change? it seems like it [won't affect existing mod builds](https://docs.oracle.com/javase/specs/jls/se17/html/jls-13.html#jls-13.4.7), but I think any addon that implements these methods would get a compilation error after this change.

assuming they didn't publish it, they'd have a compile error + linkage error at runtime. this can be fixed ahead-of-time by publishing the methods (conveniently I already do so)

Robotgiggle (approved):
As long as existing builds work alongside this I think it's fine - compilation errors when building an addon can pretty easily be fixed since you likely have the code open to build it anyway.

PR #908, opened 2025-07-15
Optimized Spell Circles

Stick404:
Replaced `knownPositions` with 2 Block Positions, so a AABB list would not need to be saved and loaded.
Made `reachedPositions` into a long; this is to remove a chunk ban and memory leak issue that happen when a Spell Circle loops for too long.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
@@ -50,19 +50,18 @@ public class CircleExecutionState {
     public CastingImage currentImage;
     public @Nullable UUID caster;
     public @Nullable FrozenPigment casterPigment;
-    // This controls the speed of the current slate
     public long reachedSlate;
 
     // Tracks the highest pos, and lowest pos of the AABB
-    public BlockPos positivePos;
-    public BlockPos negativePos;
+    public BlockPos greaterCorner;
+    public BlockPos lesserCorner;
 
     public final AABB bounds;
 
 
     protected CircleExecutionState(BlockPos impetusPos, Direction impetusDir, HashSet<BlockPos> reachedPositions,
-        BlockPos currentPos, Direction enteredFrom, CastingImage currentImage, @Nullable UUID caster,
-        @Nullable FrozenPigment casterPigment, Long reachedSlate, BlockPos positivePos, BlockPos negativePos) {
+       BlockPos currentPos, Direction enteredFrom, CastingImage currentImage, @Nullable UUID caster,
+       @Nullable FrozenPigment casterPigment, Long reachedSlate, BlockPos greaterCorner, BlockPos lesserPos) {
         this.impetusPos = impetusPos;
         this.impetusDir = impetusDir;
         this.reachedPositions = reachedPositions;
@@ -73,9 +72,9 @@ public class CircleExecutionState {
         this.casterPigment = casterPigment;
         this.reachedSlate = reachedSlate;
 
-        this.positivePos = positivePos;
-        this.negativePos = negativePos;
-        this.bounds = new AABB(positivePos,negativePos);
+        this.greaterCorner = greaterCorner;
+        this.lesserCorner = lesserPos;
+        this.bounds = new AABB(greaterCorner, lesserPos);
     }
 
     public @Nullable ServerPlayer getCaster(ServerLevel world) {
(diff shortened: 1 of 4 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -110,29 +122,34 @@ protected CircleExecutionState(BlockPos impetusPos, Direction impetusDir, Set<Bl
             }
 
             if (seenGoodPosSet.add(herePos)) {
+                lastBlockPos.set(herePos);
+                // Checks to see if it should update the most positive/negative block pos
```
abilliontrillionstars: using Math.max() here would be more readable, i.e. 
`positiveBlock.setX(Math.max( herePos.getX(), positiveBlock.getX() ));`

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -31,42 +31,51 @@ public class CircleExecutionState {
...
     // Does contain the starting impetus
-    public final Set<BlockPos> knownPositions;
-    public final List<BlockPos> reachedPositions;
+    public final HashSet<BlockPos> reachedPositions;
     public BlockPos currentPos;
     public Direction enteredFrom;
     public CastingImage currentImage;
     public @Nullable UUID caster;
     public @Nullable FrozenPigment casterPigment;
+    // This controls the speed of the current slate
```
abilliontrillionstars: variable names aren't *too* descriptive here, something better might be `slateWait`, and `greaterCorner` / `lesserCorner`

Stick404: Alright, fixed this. The reachedSlate is just a number of how many slates it has ran, so I feel like changing it to slateWait is not the greatest option? But did change the corners to be more descriptive

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
Stick404: Thanks! Made it more readable now.

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -111,33 +121,33 @@ protected CircleExecutionState(BlockPos impetusPos, Direction impetusDir, Set<Bl
...
-            return new Result.Err<>(seenGoodPositions.get(seenGoodPositions.size() - 1));
+            return new Result.Err<>(lastBlockPos);
         }
 
-        var knownPositions = new HashSet<>(seenGoodPositions);
-        var reachedPositions = new ArrayList<BlockPos>();
-        reachedPositions.add(impetus.getBlockPos());
-        var start = seenGoodPositions.get(0);
-
+        seenGoodPosSet.add(impetus.getBlockPos());
```
Robotgiggle: What exactly is the point of this line? The check on line 143 should have already ensured that the impetus pos is in this set, no?

Is this supposed to be creating a new HashSet to pass to `reachedPositions`?

Common/src/main/java/at/petrak/hexcasting/api/casting/circles/CircleExecutionState.java
```diff
@@ -147,8 +157,10 @@ protected CircleExecutionState(BlockPos impetusPos, Direction impetusDir, Set<Bl
             casterUUID = caster.getUUID();
         }
         return new Result.Ok<>(
-            new CircleExecutionState(impetus.getBlockPos(), impetus.getStartDirection(), knownPositions,
-                reachedPositions, start, impetus.getStartDirection(), new CastingImage(), casterUUID, colorizer));
+            new CircleExecutionState(impetus.getBlockPos(), impetus.getStartDirection(),
+                seenGoodPosSet, impetus.getBlockPos().offset(impetus.getStartDirection().getNormal()),
```
Robotgiggle: Why are we passing the entire set of found circle positions to the `reachedPositions` parameter of the new CircleExecutionState? Shouldn't this just contain the impetus pos and nothing else?

Robotgiggle (approved):
I fixed the initialization of the `reachedPositions` set, and also deprecated `BlockEntityAbstractImpetus.getBounds` since it's now unused. The rest seems good.

Robotgiggle (changes requested):
Never mind, there seems to be a major issue with the bounds/ambit check. The AABB is reporting very weird values when I log it, causing circles to have much larger ambit cuboids than intended.

Robotgiggle (approved):
AABB issue fixed by properly initializing the corner positions before running the loop to detect the cuboid. Should be all good now.

PR #910, opened 2025-07-15
Better Circle Scanning

Stick404:
Improves Spell Circle scanning by making a custom system to use `ServerChunkCache#getChunkFuture`'s quicker chunk loading.
Creates a Helper Class in case spells/addons wish to use this system


Common/src/main/java/at/petrak/hexcasting/api/utils/ChunkScanning.kt
```diff
@@ -0,0 +1,68 @@
...
+        return chunks.get(chunkPos).getBlockState(blockPos)
+    }
+
+    fun getBlockEntity(blockPos: BlockPos): BlockEntity? {
+        val chunkPos = ChunkPos(blockPos).toLong()
+        if (!cacheChunk(chunkPos)){
+            return null
+        }
+        return chunks.get(chunkPos).getBlockEntity(blockPos)
+    }
```
s5bug:
```suggestion
    fun cacheChunk(chunk: ChunkPos): ImposterProtoChunk? {
        val chunkLong = chunk.toLong()
        // If we have the chunk already, we can skip fetching it
        val existing = chunks.get(chunkLong)
        if (existing != null){
            return existing
        }
        val future = level.chunkSource.getChunkFuture(chunk.x,chunk.z, ChunkStatus.EMPTY,true).get()
        if (future.left().isPresent){
            val next = future.left().get() as ImposterProtoChunk
            chunks.put(chunkLong, next)
            return next
        }
        HexAPI.LOGGER.warn("Failed to get chunk at {}!",chunk)
        return null
    }

    fun cacheChunk(chunk: Long): ImposterProtoChunk? {
        return cacheChunk(ChunkPos(chunk))
    }

    fun getBlock(blockPos: BlockPos): BlockState? {
        val chunkPos = ChunkPos(blockPos).toLong()
        return cacheChunk(chunkPos)?.getBlockState(blockPos)
    }

    fun getBlockEntity(blockPos: BlockPos): BlockEntity? {
        val chunkPos = ChunkPos(blockPos).toLong()
        return cacheChunk(chunkPos)?.getBlockEntity(blockPos)
    }
```

Common/src/main/java/at/petrak/hexcasting/api/utils/ChunkScanning.kt
```diff
@@ -0,0 +1,68 @@
...
+     */
+    fun cacheChunk(chunk: ChunkPos): Boolean {
+        val chunkLong = chunk.toLong()
+        // We have the chunk already, so we can skip it
+        if (chunks.contains(chunkLong)){
+            return true
+        }
+        val future = level.chunkSource.getChunkFuture(chunk.x,chunk.z, ChunkStatus.EMPTY,true).get()
+        if (future.left().isPresent){
+            chunks.put(chunkLong, future.left().get() as ImposterProtoChunk)
```
s5bug: Why is this cast safe?

Stick404: Oh woops, I don't believe it is; thank you for the catch! I'll make `ChunkScanning` hold a list of `ChunkAccess`s instead

Robotgiggle (approved):
LGTM! Can confirm this significantly improves performance of high density circle spam, even when connected to enormous ambit rods.

PR #913, opened 2025-07-23
Update zh_cn

ChuijkYahus:
This is still under proofreading stage, and would be out from draft once that's finished.


object-Object:
thanks!

Issue #327, opened 2022-11-25
Greater teleport causes dedicated server to hang indefinitely when teleporting to unloaded chunks


Flyte-less:
Using version 0.9.4 on forge.

When using the spell long distance it can consistently cause the server to enter a frozen state, forcing every player to time out. The only way to recover from this state is by force killing the server process.

Notes: 
It seems that if the destination chunk is loaded the greater teleport will not cause a server crash, regardless of how many of the surrounding chunks are loaded as well

The built in /tp command does not seem affected.

Lowering the render and simulation distances to the very lowest seems to mitigate this issue.

[Modlist if relevant.](https://gist.github.com/Flyte-less/fd200285ebb2e4eecad501a1538b9922) No logs or crash reports to provide as the server is unable to record anything during it's frozen state.

Apologies if this isn't super helpful, I only have one dedicated server and was unable to replicate this bug in singleplayer.

Flyte-less:
One more piece of information that I hope is useful. Remember how I said that lowering the render distance seemed to help? Well it looks like doing that causes none of the surrounding chunks to appear even after setting the render distance back to normal.

![image](https://user-images.githubusercontent.com/83427235/204070838-bfbbd817-c6e3-4063-ab75-3417ea3515f0.png)

Relogging or spamming the render distance option seems to fix it though.

gamma-delta:
That looks to me like your server just doesn't have enough memory to load those chunks that quickly ... but I'm not sure how come `/tp` doesn't crash, then. Lemme take a look.

gamma-delta:
That should fix it but let me know if it doesn't.

St0rmC4st3r:
> That should fix it but let me know if it doesn't.

Letting you know that it doesn't. Playing TNP Limitless 5 2.30.0. Minecraft version: 1.18.2. Forge version: 40.2.0.
Seems to hang the server inconsistently, as in, sometimes it does unhang, and sometimes it hangs indefinitely.
I also tried claiming and loading the chunks via FTB Chunks, but this doesn't seem to affect the bug.

redstonerblast:
This bug is still occurring to me in 1.20.1

PR #916, opened 2025-07-30
Fix various issues with Blink and GTP

### LINKED ISSUE #898: Teleports Failing
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

0.11.2

### Modloader version

47.4.0

### Modpack info

Hexcasting + hexal + dependancies only

### The latest.log file

https://mclo.gs/HZGHqXA

### Issue description

Teleporting, especially while moving or falling will revert the teleport after a frame or two, with an accompanying "moved wrongly".

### Steps to reproduce

Make an artifact that teleports you somewhere.
Activate it while running and jumping around, especially off of heights
Notice it not working.

### Other information

Using the hexal and moreiotas from hexxytest

Robotgiggle:
For some reason the `teleportRespectSticky()` method, which is used by both Blink and GTP, entirely reimplements vanilla teleport functionality for players rather than just calling `ServerPlayer.teleportTo()`. This seems to have been the cause for the can't-teleport-while-moving issue, as replacing all the custom reimplementation with just the vanilla call seems to entirely fix #898 without causing any new problems on its own.

Based on the git blame, much of the manual teleport implementation was added in an attempt to solve #327, but a comment on that issue suggests that it was never actually resolved, so removing the "fixes" shouldn't be a problem.

This PR also fixes a couple other issues with OpTeleport:
- The `canChangeDimensions()` check was entirely invalidating the sticky-teleport mechanic by preventing entities anywhere in a passenger stack from even being valid targets, so it has been removed
- The check for teleport-immune passengers has been moved from cast() to execute() so it can throw a mishap rather than just silently failing and still consuming media


beholderface:
Isn't the `canChangeDimensions` method also used for things that really shouldn't be moveable?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpBlink.kt
```diff
@@ -27,9 +27,15 @@ object OpBlink : SpellAction {
         val delta = args.getDouble(1, argc)
         env.assertEntityInRange(target)
 
-        if (!target.canChangeDimensions() || target.type.`is`(HexTags.Entities.CANNOT_TELEPORT))
+        if (target.type.`is`(HexTags.Entities.CANNOT_TELEPORT))
```
object-Object, line 30: Doesn't have to be in this PR, but I think this logic should be refactored out into a common helper to be referenced by both OpBlink and OpTeleport, both to reduce duplication in our code and to make it available for use by addons.

PR #931, opened 2025-09-06
Update error logging for bad patterns in hexdoc

### LINKED ISSUE #930: Improve hexdoc error message for unknown patterns
https://github.com/FallingColors/HexMod/blob/3c5d9df75aa6d08b04bcebb0153a079aefb8793a/doc/src/hexdoc_hexcasting/book/page/pages.py#L22

If this lookup fails, the Pydantic error is very unhelpful: `KeyError: (pattern id)`. We should add a better one.

navarchus:
Fixes #930
Update error logging for bad patterns in hexdoc

### DIFF
--- a/doc/src/hexdoc_hexcasting/book/page/pages.py
+++ b/doc/src/hexdoc_hexcasting/book/page/pages.py
@@ -19,7 +19,10 @@ class LookupPatternPage(PageWithOpPattern, type="hexcasting:pattern"):
     @model_validator(mode="after")
     def _post_root_lookup(self, info: ValidationInfo):
         hex_ctx = HexContext.of(info)
-        self._patterns = [hex_ctx.patterns[self.op_id]]
+        try:
+            self._patterns = [hex_ctx.patterns[self.op_id]]
+        except KeyError as e:
+            raise KeyError(e.add_note("Unknown Pattern ID!"))
         return self
 
     @model_validator(mode="after")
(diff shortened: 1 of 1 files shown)


doc/src/hexdoc_hexcasting/book/page/pages.py
```diff
@@ -19,7 +19,10 @@ def patterns(self) -> list[PatternInfo]:
     @model_validator(mode="after")
     def _post_root_lookup(self, info: ValidationInfo):
         hex_ctx = HexContext.of(info)
-        self._patterns = [hex_ctx.patterns[self.op_id]]
+        try:
+            self._patterns = [hex_ctx.patterns[self.op_id]]
+        except KeyError as e:
+            raise KeyError(e.add_note("Unknown Pattern ID!"))
```
object-Object: I'd prefer to check the condition ahead of time rather than catch and re-throw. Something like this:
```suggestion
        pattern = hex_ctx.patterns.get(self.op_id)
        if pattern is None:
            raise ValueError(f"Unknown pattern ID (check your pattern stubs in hexdoc.toml): {self.op_id}")
        self._patterns = [pattern]
```
I also changed the exception type to ValueError since that's what Pydantic [recommends](https://docs.pydantic.dev/latest/concepts/validators/#raising-validation-errors) for validation errors.

PR #933, opened 2025-09-06
Change mishap parent class to `RuntimeException` to prevent game crashes

### LINKED ISSUE #891: Throwing a mishap in `RenderedSpell.cast` crashes the game
## Summary

Mishaps thrown in side effects, including `AttemptSpell`, are not caught by the VM and crash the game/server. For example: https://mclo.gs/6jEWQAo

This was supposed to be fixed by #630, but that PR missed the fact that `Mishap` extends `Throwable`, not `Exception`, so the following code does *not* catch mishaps:

https://github.com/FallingColors/HexMod/blob/5da1bf82464d82aa303020e3fdf1fca17b9055d1/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt#L75-L80

This is a pretty common mistake in addons, especially since I don't think we've really *documented* that you're not supposed to throw mishaps in `RenderedSpell` subclasses.

## Potential solutions

The simple fix would be to add another catch block specifically for mishaps. We should also document that mishaps should not be thrown in `RenderedSpell.cast` implementations.

However, I think the proper fix would be to change the base class of `Mishap` to `Exception` or `RuntimeException`. I don't really see a valid reason for mishaps not to be exceptions. As it currently stands, to catch mishaps, you either need to catch `Throwable`, which is [bad practice](https://stackoverflow.com/a/498226) since that [also includes](https://kotlinlang.org/docs/exceptions.html#exception-hierarchy) things like `OutOfMemoryError` and `StackOverflowError`; or specifically catch `Mishap` along with / instead of `Exception`, which is why this issue occurred in the first place.

navarchus:
Fixes #891
Change mishap parent class to `RuntimeException`

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
@@ -74,7 +74,7 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
             lastResolutionType = image2.resolutionType
             try {
                 performSideEffects(image2.sideEffects)
-            } catch (e: Exception) {
+            } catch (e: RuntimeException) {
                 e.printStackTrace()
                 performSideEffects(listOf(OperatorSideEffect.DoMishap(MishapInternalException(e), Mishap.Context(null, null))))
             }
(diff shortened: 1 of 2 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -74,7 +74,7 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
             lastResolutionType = image2.resolutionType
             try {
                 performSideEffects(image2.sideEffects)
-            } catch (e: Exception) {
+            } catch (e: RuntimeException) {
```
object-Object: This should stay as Exception. We want to catch all exceptions here, to prevent inadvertently-thrown exceptions from escaping the VM and crashing the entire server.

PR #935, opened 2025-09-06
Allow action and spell action to throw mishaps

### LINKED ISSUE #874: Add `@Throws(Mishap::class)` to `Action#operate` and similar subclass methods
### Describe the feature

As per the [Kotlin docs](https://kotlinlang.org/docs/java-to-kotlin-interop.html#checked-exceptions), Kotlin code needs to use the `@Throws` annotation to allow Java code to throw/catch exceptions in some cases. Adding `@Throws(Mishap::class)` to `Action#operate`, `SpellAction#execute`, etc would make it easier for Java addons to throw mishaps without resorting to [scuffed workarounds](https://github.com/SamsTheNerd/HexGloop/blob/c2b5819a2c76e8a9f54c81807873e80066386ec6/common/src/main/java/com/samsthenerd/hexgloop/casting/OpSetLabel.java#L63).

Note that we apparently already did this for `Operator` and `OperatorBasic`:

https://github.com/FallingColors/HexMod/blob/ef2cd28b2a07782cab215db309b1c7acabeec7fb/Common/src/main/java/at/petrak/hexcasting/api/casting/arithmetic/operator/Operator.kt#L29-L30

### Additional context

_No response_

navarchus:
Fixes #874
allow action and spell action to throw mishaps

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/Action.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/Action.kt
@@ -5,6 +5,7 @@ import at.petrak.hexcasting.api.casting.eval.OperationResult
 import at.petrak.hexcasting.api.casting.eval.vm.CastingImage
 import at.petrak.hexcasting.api.casting.eval.vm.SpellContinuation
 import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.casting.mishaps.Mishap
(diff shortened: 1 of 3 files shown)


navarchus:
unsure if this is everything, let me know if anything else should be changed

object-Object (changes requested):
Can you add it to ConstMediaAction as well?

object-Object (approved):

Issue #950, opened 2025-10-10
Add small type blurb to iota expected mishaps


miyucomics:
### Describe the feature

Mishaps should write a little blurb first, like `Impulse expected an entity at index 1 of the stack, but got a list [Creeper]`. A lot of people have been getting confused about entities and a lone entity in a list, likely because they glance over the brackets or think it's for decor. Just adding a quick clarification would help UX quite a bit. The way I envision this being implemented is a custom translation key for all iota types, something like `"iota.hexcasting:null": "null"` and `"iota.hexcasting:vec3": "a vector"`.

PR #951, opened 2025-10-10
More informative mishaps

Robotgiggle:
- Implements #950 
- Changes the invalid-pattern mishap to actually display the offending pattern
image
image

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidPattern.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidPattern.kt
@@ -10,7 +10,10 @@ import at.petrak.hexcasting.api.pigment.FrozenPigment
 import net.minecraft.network.chat.Component
 import net.minecraft.world.item.DyeColor
 
-class MishapInvalidPattern(val pattern: HexPattern? = null) : Mishap() {
+class MishapInvalidPattern(val pattern: HexPattern?) : Mishap() {
+    @Deprecated("Provide the pattern that caused the mishap as an argument")
+    constructor() : this(null) {}
+
     override fun accentColor(ctx: CastingEnvironment, errorCtx: Context): FrozenPigment =
         dyeColor(DyeColor.YELLOW)
 
@@ -22,6 +25,6 @@ class MishapInvalidPattern(val pattern: HexPattern? = null) : Mishap() {
 
     override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context): Component? {
         if (pattern == null) return error("invalid_pattern_generic")
-        else return error("invalid_pattern", PatternIota.display(pattern))
+        return error("invalid_pattern", PatternIota.display(pattern))
     }
 }


Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidIota.kt
```diff
@@ -23,11 +24,14 @@ class MishapInvalidIota(
         stack[stack.size - 1 - reverseIdx] = GarbageIota();
     }
 
-    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context) =
-        error(
+    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context): Component? {
+        val typeStr = HexIotaTypes.REGISTRY.getKey(perpetrator.getType())?.getPath();
```
object-Object: Not a huge fan of this; what if two mods add iota types with the same path?

Robotgiggle: It's currently set up this way because the `hexcasting.mishap.invalid_value.class.<type>` lang keys already exist (for use in the old version of MishapInvalidIota) with exactly the string that's needed for this part of the message, and are already used by basically every addon. The alternative would be to make a completely new set of lang keys for this based on the full ResourceLocation, but that would A) require writing out effectively the exact same strings twice for every iota type and B) require every addon to update their lang to do so as well in order to support this version of the base mod.

I agree that it's not a perfect solution, but I can't really think of a better one that doesn't introduce other issues as described above. If you have one in mind though, please let me know!

On another note, two addons that use the same path will likely run into issues regardless of this change, since using the MishapInvalidIota.ofType() method requires defining a `hexcasting.mishap.invalid_value.class.<type>` lang key for your iota type, and those will conflict with each other anyway unless one of them uses a type name in the lang that isn't the same as the identifier path.

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -1025,15 +1053,17 @@
...
           double: "a number",
           boolean: "a boolean",
-          vector: "a vector",
+          vec3: "a vector",
           list: "a list",
-          widget: "an influence",
           pattern: "a pattern",
+          continuation: "a jump iota",
+          garbage: "garbage",
+          null: "null",
```
object-Object, line 1076: Are these changes still necessary with the new key structure?

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidPattern.kt
```diff
@@ -4,10 +4,12 @@ import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
 import at.petrak.hexcasting.api.casting.eval.ResolvedPatternType
 import at.petrak.hexcasting.api.casting.iota.GarbageIota
 import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.casting.iota.PatternIota
+import at.petrak.hexcasting.api.casting.math.HexPattern
 import at.petrak.hexcasting.api.pigment.FrozenPigment
 import net.minecraft.world.item.DyeColor
 
-class MishapInvalidPattern : Mishap() {
+class MishapInvalidPattern(val pattern: HexPattern) : Mishap() {
```
object-Object: This is a breaking change. Do any addons throw this mishap? If so, there should be a fallback constructor taking no arguments.

PoolloverNathan: What would be displayed via the fallback constructor? No / placeholder pattern?

PoolloverNathan: GitHub search shows no matches except for HexMod and forks thereof, so I doubt that this would break addons. However, it is technically ABI-breaking, so some fallback would be safest.

object-Object: Probably just the old message?

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
Robotgiggle: Strictly speaking they aren't necessary, but I do think most of them are still good to have for purposes of clarity (there was previously no way to specify that something was expecting a jump iota, for instance). I could move them to a separate PR if you want, but it seems like a really small change to make an entire PR for.

I can definitely change `vec3` back to `vector` though, to avoid breaking any addons that might be using the old syntax.

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidPattern.kt
```diff
@@ -4,10 +4,13 @@ import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
...
 import at.petrak.hexcasting.api.casting.iota.GarbageIota
 import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.casting.iota.PatternIota
+import at.petrak.hexcasting.api.casting.math.HexPattern
 import at.petrak.hexcasting.api.pigment.FrozenPigment
+import net.minecraft.network.chat.Component
 import net.minecraft.world.item.DyeColor
 
-class MishapInvalidPattern : Mishap() {
+class MishapInvalidPattern(val pattern: HexPattern? = null) : Mishap() {
```
object-Object: Can this be a separate constructor annotated with `@Deprecated`, instead of an optional argument?

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidPattern.kt
```diff
@@ -17,6 +20,8 @@ class MishapInvalidPattern : Mishap() {
         stack.add(GarbageIota())
     }
 
-    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context) =
-        error("invalid_pattern")
+    override fun errorMessage(ctx: CastingEnvironment, errorCtx: Context): Component? {
+        if (pattern == null) return error("invalid_pattern_generic")
+        else return error("invalid_pattern", PatternIota.display(pattern))
```
object-Object: Nitpick: this `else` is unnecessary

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInvalidPattern.kt
Robotgiggle: Done

object-Object (approved):

Robotgiggle:
Hmm yeah that's a good idea, totally forgot that mishap existed - I'll make a new PR for that sometime soon.

Robotgiggle:
Created #970 to handle that. Note that MishapDisallowedAction is actually based on the _action_, not the pattern that triggered it, so the display I made for it is a little different.

PR #957, opened 2025-10-25
Explicitly change PatternDrawState visibility to public

tizu69:


Common/src/main/java/at/petrak/hexcasting/client/gui/GuiSpellcasting.kt
```diff
@@ -508,15 +508,15 @@ class GuiSpellcasting constructor(
...
-        object BetweenPatterns : PatternDrawState()
+        public object BetweenPatterns : PatternDrawState()
 
         /** We just started drawing and haven't drawn the first line yet. */
-        data class JustStarted(val start: HexCoord) : PatternDrawState()
+        public data class JustStarted(val start: HexCoord) : PatternDrawState()
 
         /** We've started drawing a pattern for real. */
-        data class Drawing(val start: HexCoord, var current: HexCoord, val wipPattern: HexPattern) : PatternDrawState()
+        public data class Drawing(val start: HexCoord, var current: HexCoord, val wipPattern: HexPattern) : PatternDrawState()
```
object-Object, line 519: None of these should be necessary - just removing the `private` modifier on `PatternDrawState` (not explicitly setting anything to `public`) should be adequate afaik.

PR #959, opened 2025-10-28
Check the center of BlockPos for ambit instead of the corner

### LINKED ISSUE #946: `CastingEnvironment#assertPosInRange` should check the center of the block instead of the corner
This is the current implementation of `assertPosInRange` and `assertPosInRangeForEditing`:

https://github.com/FallingColors/HexMod/blob/871f9387a3e1ccf0231a3e90c31e5d8472d46fde/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java#L332-L340

`vec.getX(), vec.getY(), vec.getZ()` returns the lower-north-west corner of the block. This is arguably very confusing behaviour. Raycasts return the center of the block, so the most intuitive way to check if a pattern will have ambit (subtracting from your position) may fail in some edge cases.

Additionally, this has a nasty interaction with `canEditBlockAt`, because that method checks the *center* of the block:

https://github.com/FallingColors/HexMod/blob/871f9387a3e1ccf0231a3e90c31e5d8472d46fde/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java#L342-L344

This means that when using `assertPosInRangeForEditing` for certain positions on the very edge of your ambit, the `assertVecInRange` check succeeds but the `canEditBlockAt` check fails, resulting in the "forbidden to you" mishap being incorrectly thrown instead of the "out of range" mishap.

To fix this, we should just make `assertPosInRange` and `assertPosInRangeForEditing` check if `Vec3.atCenterOf(vec)` is in range, like `canEditBlockAt` already does.

### LINKED ISSUE #958: Misjudgement of `no breaking permission` in `CastingEnvironment`
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

github latest

### Modloader version

not relevant

### Modpack info

_No response_

### The latest.log file

_No response_

### Issue description

in function `assertPosInRangeForEditing`
https://github.com/FallingColors/HexMod/blob/285c9df77a036a48c96f1c153e90401d386ffc7b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java#L336-L344
it first checks the corner vector of certain block position to throw `too_far` exception, then checks `Vec3.atCenterOf(vec)` center vector again inside `canEditBlockAt`, this time without throwing; therefore, if a block's flooring position is covered in ambit while center not, it will be judged as `forbidden` instead of `too_far`, even in singleplayer without permission mods

### Steps to reproduce

produced in someone else's floodfill dig spell, located by code review

### Other information

btw is it welcomed if I make a PR directly when spotting things like this :3

YukkuriC:
not sure if the `isVecInRange` inside `canEditBlockAt` could be removed safely `_(:з」∠)_`
Fixes #946. Fixes #958.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
@@ -330,13 +330,15 @@ public abstract class CastingEnvironment {
     }
 
     public final void assertPosInRange(BlockPos vec) throws MishapBadLocation {
-        this.assertVecInRange(new Vec3(vec.getX(), vec.getY(), vec.getZ()));
+        Vec3 centered = Vec3.atCenterOf(vec);
+        this.assertVecInRange(centered);
     }
 
     public final void assertPosInRangeForEditing(BlockPos vec) throws MishapBadLocation {
-        this.assertVecInRange(new Vec3(vec.getX(), vec.getY(), vec.getZ()));
+        Vec3 centered = Vec3.atCenterOf(vec);
+        this.assertVecInRange(centered);
         if (!this.canEditBlockAt(vec))
-            throw new MishapBadLocation(Vec3.atCenterOf(vec), "forbidden");
+            throw new MishapBadLocation(centered, "forbidden");
     }
 
     public final boolean canEditBlockAt(BlockPos vec) {


object-Object:
> not sure if the `isVecInRange` inside `canEditBlockAt` could be removed safely

If it were `private` I'd say remove it, but since it's `public` it's possible that it was used by some addon without the preceding check, so I think we should leave it.

Issue #926, opened 2025-08-20
Iota.display() is called on both client and server, so client configs can't consistently affect it


Robotgiggle:
Iota.display() can be called server-side by mishaps and Reveal, and on Fabric the server does not have a client config. This means that any client config settings that would affect how iotas are displayed have to be ignored on server-side display() calls to avoid a NullPointerException, which leads to inconsistency in how things are displayed.

Robotgiggle:
Closing this since the behavior causing this "issue" was removed in #1113, and it probably shouldn't have been implemented in the first place.

PR #965, opened 2025-11-18
queueExecuteAndWrapIotas: allow specifying continuation

PoolloverNathan:
Adds a defaulted queueExecuteAndWrapIotas() parameter to allow specifying the continuation (in addition to the generated `FrameEvaluate` frame). This should not change current behavior, and, due to `@JvmOverloads`, does not break compatibility with existing addons.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
@@ -41,7 +41,8 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
     @JvmOverloads
     fun queueExecuteAndWrapIotas(iotas: List<Iota>, world: ServerLevel, nextContinuation: SpellContinuation = SpellContinuation.Done): ExecutionClientView {
         // Initialize the continuation stack to a single top-level eval for all iotas.
-        var continuation = nextContinuation.pushFrame(FrameEvaluate(SpellList.LList(0, iotas), false))
+        // HACK: Ideally, we'd have a separate (SpellContinuation, ServerWorld) overload, but then 0.11.3 would cause mixins to break. Look into properly splitting this in 1.21?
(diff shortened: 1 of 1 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -38,9 +38,11 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
      *
      * Mutates this
      */
-    fun queueExecuteAndWrapIotas(iotas: List<Iota>, world: ServerLevel): ExecutionClientView {
+    @JvmOverloads
+    fun queueExecuteAndWrapIotas(iotas: List<Iota>, world: ServerLevel, nextContinuation: SpellContinuation = SpellContinuation.Done): ExecutionClientView {
         // Initialize the continuation stack to a single top-level eval for all iotas.
-        var continuation = SpellContinuation.Done.pushFrame(FrameEvaluate(SpellList.LList(0, iotas), false))
+        // HACK: Ideally, we'd have a separate (SpellContinuation, ServerWorld) overload, but then 0.11.3 would cause mixins to break. Look into properly splitting this in 1.21?
```
object-Object, line 44: Can you open an issue to track this?

PoolloverNathan: sure, what should it look like (or do you have examples)? not exactly sure what the issue conventions are

object-Object: Something along the lines of https://github.com/FallingColors/HexMod/issues/926 would be fine. Just a reminder to clean up the code the next time we can make a breaking change.

PoolloverNathan:
seems to break hexxyinthealps and all mods that inject into this method weirdly, https://mclo.gs/rn9Glcl

PR #968, opened 2025-11-20
Revert #965

object-Object:
This reverts commit c64e5d7bc8c9872edd42daa420b29a3649997eda, reversing changes made to 22b7a6f4d00eaafa9b5b83748d5c5b165e80d3a6.

The changes made in #965 were meant to avoid any impact for addons, but we've since realized that adding an optional argument to a method still breaks mixins targeting that method. This caused a crash with HexxyInTheAlps and may also break other addons.

While mixin impacts aren't technically breaking changes, we still don't really want to break an unknown number of addons this close to a release, so we'll resolve this issue by reverting #965 and revisiting this change at a later date.


PoolloverNathan (approved):
Don't see any issues with this; when's our next breaking change window?

PR #969, opened 2025-11-20
Revert "Revert #965"

### LINKED ISSUE #967: PR #965 breaks mixins
The changes in #965 can result in mixin application failures: https://mclo.gs/rn9Glcl

While mixin impacts aren't technically breaking changes, we still don't really want to break an unknown number of addons this close to a release, so we'll resolve this issue by reverting #965 and revisiting this change at a later date.

cc @PoolloverNathan

PoolloverNathan:
Reverts #968, reapplying #965, resolves #967. Per a GitHub search, it seems that the mixin conflict in question is [unique to Hexxy Searches For a Lost Cat in the Alps](https://github.com/search?q=%40WrapMethod+queueExecuteAndWrapIotas&type=code). The behavior of a single addon, especially one that is only used on one server, is not our responsibility, and alternatives to introducing a parameter are much hackier than they need to be.


PoolloverNathan:
I can probably just use a mixin for this

PR #976, opened 2025-12-11
[1.20.1] `CastingEnvironment#getCasterPosition`

TechTastic:
I had a slight issue with implementing Hexcasting compatibilty for Ambit and Range testing in that i couldnt cleanly get the casting position from `CastingEnvironment`

This PR's purpose is to add such a method and its base implementations to Hexcasting

This *is* a breaking change for those addons that make custom `CastingEnvironment` implementations

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
@@ -129,6 +129,13 @@ public abstract class CastingEnvironment {
     public abstract LivingEntity getCastingEntity();
 
     /**
+     * Gets the caster's position. Useful for compatibility.
+     *
+     * @return the position of the caster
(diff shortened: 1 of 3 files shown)


miyucomics:
Isn't there a mishap position method that works similarly? We could consolidate both and have it under `getCastingPosition`, for impeti and other sources that are not linked to a casting entity, and mishap particles would also spawn from there.

object-Object:
> Isn't there a mishap position method that works similarly? We could consolidate both and have it under `getCastingPosition`, for impeti and other sources that are not linked to a casting entity, and mishap particles would also spawn from there.

The mishap position and "caster position" aren't necessarily the same. For example, circles would likely set the caster position to the impetus, while the mishap position is currently the slate that triggered the mishap.

c-Caelum:
With Kotlin's ability to fake methods in classes, you could make this yourself with a fraction of the breakage.

PR #984, opened 2025-12-24
Actually stop using AWT

vgskye:
https://github.com/openjdk/jdk/blob/jdk-21-ga/src/java.desktop/share/classes/java/awt/Color.java#L268-L281

fixes a possible bug on macOS


vgskye:
> There's another usage of AWT here, do we need to remove that too?
> 
> https://github.com/FallingColors/HexMod/blob/7202ed7206ebd13682c72821b88b0f36b68ff36d/Common/src/main/java/at/petrak/hexcasting/client/render/PatternTextureManager.java#L12

nope- awt.geom is safe iirc

object-Object (approved):

PR #987, opened 2025-12-27
Exploit fixes for media_consumption attribute

### LINKED ISSUE #983: Media Consumption Modifier attribute affects non-player proxy casters
### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

The newly added `media_consumption_modifier` attribute applies to any cast where `getCastingEntity()` isn't null. This means that if a player has a modified media consumption value, the modification will also apply to spell circles that they activate or are bound to, and to Hexal wisps that they own.

This is rather unintuitive - why should a player attribute have any effect on the media consumption of automated constructs like circles? Furthermore, it can allow media duping when combined with addon patterns like [Export Media](https://hexical.hexxy.media/v/latest/main/en_us/#patterns/spells/circle_spells@hexical:create_dust) that extract or transfer media from proxy casters.

I encountered this exact problem when adding a media discount effect to Hierophantics (https://github.com/Robotgiggle/hierophantics/issues/11) and solved it by only applying the discount if the current environment is a PlayerBasedCastEnv. I would recommend a similar solution here.

### Steps to reproduce

1. Apply a media discount to yourself using the `media_consumption_modifier` attribute
2. Observe that the discount also applies to spell circles that you activate
3. Observe that the discount also applies to Hexal wisps that you own

Robotgiggle:
- Fixes #983 
- Adds a tag (`cannot_modify_cost`) for patterns that should not have their costs modified

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
-- Added the `cannot_modify_cost` tag for patterns that should ignore the `media_consumption` attribute when calculating cost, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987)
+- Added the `cannot_modify_cost` tag for patterns that should ignore the `media_consumption` attribute when calculating cost, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987).
 
 ### Changed
 
-- Changed the `media_consumption` attribute to only apply to player-based casting, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987)
+- Changed the `media_consumption` attribute to only apply to player-based casting, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987).
 
 ### Fixed
 


CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
...
 
+## `0.11.4` - ???
+
+### Added
+
+- Added the `cannot_modify_cost` tag for patterns that should ignore the `media_consumption` attribute when calculating cost, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987)
+
+### Changed
+
+- Changed the `media_consumption` attribute to only apply to player-based casting, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987)
```
object-Object:
```suggestion
## [UNRELEASED]

### Added

- Added the `cannot_modify_cost` tag for patterns that should ignore the `media_consumption` attribute when calculating cost, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987).

### Changed

- Changed the `media_consumption` attribute to only apply to player-based casting, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987).
```

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -241,14 +242,18 @@ public boolean isEnlightened() {
...
+
     /**
      * Attempt to extract the given amount of media. Returns the amount of media left in the cost.
      * <p>
      * If there was enough media found, it will return less or equal to zero; if there wasn't, it will be
      * positive.
      */
     public long extractMedia(long cost, boolean simulate) {
-        if (this.getCastingEntity() != null){
+        if (this instanceof PlayerBasedCastEnv
```
object-Object: The parent class having to know about a specific subclass feels like a code smell - is there another way we could implement this? Maybe something like `preExtractMediaEnvironment`?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -241,14 +242,18 @@ public boolean isEnlightened() {
         return false;
     }
 
+    public boolean costModifierAllowed = true;
```
object-Object: This should be `protected` or `private`.

Common/src/main/java/at/petrak/hexcasting/api/mod/HexTags.java
```diff
@@ -86,6 +86,11 @@ public static final class Actions {
          */
         public static final TagKey<ActionRegistryEntry> CAN_START_ENLIGHTEN = create("can_start_enlighten");
 
+        /**
+         * Actions that should not be affected by the media_consumption attribute
+         */
```
object-Object, line 91: thanks for adding documentation :)

Common/src/main/java/at/petrak/hexcasting/datagen/tag/HexActionTagProvider.java
```diff
@@ -35,6 +35,14 @@ protected void addTags(HolderLookup.Provider provider) {
             tag(ersatzActionTag(HexTags.Actions.PER_WORLD_PATTERN)).add(key);
         }
         // deciding that akashic write can be just a normal spell (as a treat)
+
+        for (var actionID : new String[]{
+            "hexal:wisp/summon/projectile", "hexal:wisp/summon/ticking", "hexical:conjure_gummy", "hexical:charm", 
+            "hexical:recharge_lamp", "lapisworks:deposit", "lapisworks:imbue_lap", "yaha:time_bomb"
```
object-Object: I'm actually not convinced that we should do this. Ideally it should be up to the addon authors to tag their patterns - addons are able to release changes much faster than we can.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/PatternIota.java
```diff
@@ -94,9 +94,16 @@ public boolean toleratesOther(Iota that) {
                     // this gets caught down below
                     throw new MishapUnenlightened();
                 }
+
+                if (isOfTag(IXplatAbstractions.INSTANCE.getActionRegistry(), key, HexTags.Actions.CANNOT_MODIFY_COST)) {
+                    vm.getEnv().costModifierAllowed = false;
+                } else {
+                    vm.getEnv().costModifierAllowed = true;
+                }
```
object-Object: Could we do this in `precheckAction` instead?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -241,16 +251,16 @@ public boolean isEnlightened() {
         return false;
     }
 
+    protected double costModifier = 1.0;
```
object-Object: Should be `private`, I don't think subclasses would have any reason to directly modify this

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
```diff
@@ -186,11 +187,20 @@ public <T extends CastingEnvironmentComponent> T getExtension(@NotNull CastingEn
...
+            throw new MishapDisallowedSpell("disallowed", loc);
         }
+
+        costModifier = this.checkCostModifier(loc);
+    }
+
+    /**
+     * Casting env subclasses can override this to modify the cost for a given action
+     */
+    protected double checkCostModifier(ResourceLocation loc) {
```
object-Object: Can we pass the full `PatternShapeMatch` object in? I realize that means we'd then be calling `actionKey` twice, but I think that's fine.

Also nit: can we call this `getCostModifier` instead?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedCastEnv.java
```diff
@@ -62,6 +65,14 @@ public ServerPlayer getCaster() {
         return this.caster;
     }
 
+    @Override
+    protected double checkCostModifier(ResourceLocation loc) {
+        if (this.getCastingEntity() != null && !isOfTag(IXplatAbstractions.INSTANCE.getActionRegistry(), loc, HexTags.Actions.CANNOT_MODIFY_COST)) {
```
object-Object: I don't think we need the null check anymore, we should be able to just use `caster` directly

object-Object (changes requested):

object-Object (approved):
LGTM, thanks!

PR #988, opened 2025-12-27
add method to omit commas in IotaType

TheRobbie73:
so that custom iotas may omit commas when displayed in list iotas 🥺

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
@@ -132,12 +132,13 @@ public class ListIota extends Iota {
                 out.append(IotaType.getDisplay(csub));
 
                 // only add a comma between 2 non-patterns (commas don't look good with Inline patterns)
-                // TODO: maybe add a method on IotaType to allow it to opt out of commas?
                 if (i < list.size() - 1) {
-                    var thisIotaNeedsComma = IotaType.getTypeFromTag(csub) != PatternIota.TYPE;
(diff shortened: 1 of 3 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
```diff
@@ -132,12 +132,13 @@ public Component display(Tag tag) {
...
                 if (i < list.size() - 1) {
-                    var thisIotaNeedsComma = IotaType.getTypeFromTag(csub) != PatternIota.TYPE;
-                    var nextIotaNeedsComma = IotaType.getTypeFromTag(HexUtils.downcast(list.get(i+1), CompoundTag.TYPE)) != PatternIota.TYPE;
+                    var thisType = IotaType.getTypeFromTag(csub);
+                    var nextType = IotaType.getTypeFromTag(HexUtils.downcast(list.get(i+1), CompoundTag.TYPE));
+                    var thisIotaOmitsComma = thisType != null && thisType.omitCommas();
+                    var nextIotaOmitsComma = nextType != null && nextType.omitCommas();
                     var alwaysShowCommas = HexConfig.client() != null && HexConfig.client().alwaysShowListCommas();
-                    if (thisIotaNeedsComma || nextIotaNeedsComma || alwaysShowCommas)
+                    if (!thisIotaOmitsComma || !nextIotaOmitsComma || alwaysShowCommas)
```
object-Object: This double-negative is pretty confusing. Can we switch the logic back around? eg:

```java
var thisIotaNeedsComma = this.type == null || !thisType.omitCommas();
```

Robotgiggle (approved):
Removed the double negative based on object's suggestion - LGTM

PR #992, opened 2026-01-11
Fix scroll crashloop when per-world order hasn't been calculated yet

### LINKED ISSUE #991: Grabbing a creative scroll for a newly added great spell before running recalcPatterns causes crash loop
### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

When a creative-mode Ancient Scroll tries to load the stroke order for its spell, it checks to make sure the spell exists (on line 176) but it doesn’t check to make sure that the per-world pattern for that spell has actually been calculated. If it hasn’t (ie in any situation where recalcPatterns would be necessary) then the `getCanonicalStrokesPerWorld` call on line 182 returns null, causing the `serializeToNBT` call on the following line to crash the game. Since the scroll attempts to load its pattern every inventory tick, this effectively locks you out of the world unless you edit your player data to remove the problematic scroll.

### Steps to reproduce

1. Create a world
2. Update or install an addon such that `/hexcasting recalcPatterns` is necessary
5. Grab the Ancient Scroll for the new spell from the creative tab
6. Observe that placing it in your inventory causes a crash
7. Observe that rejoining the world instantly crashes you again### Other information

The fix for this would just be to add a second null-check after the `getCanonicalStrokesPerWorld` check, just like the existing one for the `opID` value.

Robotgiggle:
Fixes #991 by adding a null check for the per-world stroke order in ItemScroll's inventory tick method.

If the new null check is triggered, the scroll will be wiped and a warning will be added to its tooltip to tell the player how to fix the problem.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
-## `0.11.4` - ???
+## [UNRELEASED]
 
 ### Fixed
 
-- Fixed a crash loop when trying to generate a creative-mode ancient scroll for a Great Spell whose per-world pattern hasn't been calculated yet, by Robotgiggle in [992](https://github.com/FallingColors/HexMod/pull/992)
+- Fixed a crash loop when trying to generate a creative-mode ancient scroll for a Great Spell whose per-world pattern hasn't been calculated yet, by Robotgiggle in [992](https://github.com/FallingColors/HexMod/pull/992).
 
 ## `0.11.3` - 2025-11-22
 


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
+## `0.11.4` - ???
+
+### Fixed
+
+- Fixed a crash loop when trying to generate a creative-mode ancient scroll for a Great Spell whose per-world pattern hasn't been calculated yet, by Robotgiggle in [992](https://github.com/FallingColors/HexMod/pull/992)
```
object-Object:
```suggestion
## [UNRELEASED]

### Fixed

- Fixed a crash loop when trying to generate a creative-mode ancient scroll for a Great Spell whose per-world pattern hasn't been calculated yet, by Robotgiggle in [992](https://github.com/FallingColors/HexMod/pull/992).
```

object-Object (approved):
LGTM

PR #1002, opened 2026-02-18
Deprecate existing matchPattern method

beholderface:


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedCastEnv.java
```diff
@@ -192,6 +192,10 @@ protected long extractMediaFromInventory(long costLeft, boolean allowOvercast, b
     }
 
     protected boolean canOvercast() {
+        //bad deployer, no more infinite bloodcasting for you
+        if (this.caster.getClass() != ServerPlayer.class){
```
object-Object: This change seems unrelated (and the commit is from a year ago), did something weird happen with the commit history?

Common/src/main/java/at/petrak/hexcasting/common/casting/PatternRegistryManifest.java
```diff
@@ -90,6 +90,7 @@ public static Pair<SpecialHandler, ResourceKey<SpecialHandler.Factory<?>>> match
      *                                      order
      *                                      for a per-world pattern.
      */
+    @Deprecated
```
object-Object: Can you add a description to the deprecated annotation, and update the javadoc to mention the new overload?

Common/src/main/java/at/petrak/hexcasting/common/casting/PatternRegistryManifest.java
```diff
@@ -119,6 +120,10 @@ public static PatternShapeMatch matchPattern(HexPattern pat, CastingEnvironment
         return new PatternShapeMatch.Nothing();
     }
 
+    public static PatternShapeMatch matchPattern(HexPattern pattern, CastingEnvironment environment){
```
object-Object, line 125: Can you add a javadoc comment to this method?

object-Object (changes requested):

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedCastEnv.java
beholderface: ack I thought that got deleted when I merged with the more recent commits

Common/src/main/java/at/petrak/hexcasting/common/casting/PatternRegistryManifest.java
object-Object: Sorry, can you also add `since=` to the deprecated annotation? I should have been more clear about what I was asking

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Deprecated.html#since()

object-Object (approved):
LGTM

PR #1007, opened 2026-02-26
Improve hexdoc handling of duplicate patterns when allow_duplicates is enabled

object-Object:
In #911, we added an `allow_duplicates` option to `[extra.hexcasting]` in `hexdoc.toml`. The goal of this feature was to allow users the option to prevent their hexdoc build from failing if they add a pattern with a duplicate signature (mainly for HexBug). However, the implementation in #911 is flawed: duplicate patterns are not added to `self.patterns`, resulting in a later validation failure for pages that use the duplicate patterns.

This PR changes the logic as follows:
- Always raise an exception if a duplicate pattern ID is detected, since there's no valid reason for that to happen and we need pattern IDs to be unique in order to look up patterns for `hexcasting:pattern` pages
- If a duplicate pattern signature is detected and `allow_duplicates` is enabled, still add the pattern to `self.patterns` so we can look it up when loading pattern pages
- Improve the formatting of the error messages for duplicate patterns

These changes have been tested in HexBug and seem to work fine.


Robotgiggle (approved):
LGTM

PR #985, opened 2025-12-24
Working 1.21 port (for the most part)

SuperKnux:

PR #1009, opened 2026-03-07
Full port to 1.21.1 neoforge of the latest version

jhyygoyhphngbhlghghg:
Sorry for huge and messy commit, was working on it on local git branch.


Common/src/main/java/at/petrak/hexcasting/common/casting/actions/SpecialHandlerCompat.kt
```diff
@@ -0,0 +1,47 @@
...
+            } else {
+                null
+            }
+        }
+    }
+}
+
+/**
+ * dare to fucking snitch on me, bitch, and i wont port this mod to other versions
+ */
```
poolcritter, line 47: You are aware that PRs are publicly visible, correct?

poolcritter (changes requested):

object-Object:
Given #985 already exists and has had significantly more work put into it, and for some other reasons that should be obvious, I don't think we'll be accepting this.

SamsTheNerd:
no i think the creative mode spell is a great feature idea actually, we should consider that one

ChloeTax:
could this pr be accepted if https://github.com/jhyygoyhphngbhlghghg/HexMod/pull/1 was merged first?

PR #1010, opened 2026-03-07
Assorted guidebook & mishap improvements

### LINKED ISSUE #977: Mishap for "no placeable block in hotbar" is both untranslated and misleading
### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

If no placeable item can be found in your hotbar, Place Block will throw `MishapBadOffhandItem.of(ItemStack.EMPTY, "placeable")`. The lang key this generates, `hexcasting.mishap.bad_item.placeable`, does not exist, leading to the following mishap message:
Image
Furthermore, even if the lang key _did_ exist, this message would still be misleading. It claims that the spell needed something in the offhand that wasn't present, but Place Block checks through your entire hotbar, not just the offhand.

### Steps to reproduce

1. Have no placeable items in your hotbar
2. Try to cast Place Block
3. Observe broken mishap message

### LINKED ISSUE #989: Caster's Glamour claims to cost 1 dust but is actually free
### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

The book entry for Caster's Glamour says that it costs about 1 amethyst dust. However, the actual cost (as seen [here](https://github.com/FallingColors/HexMod/blob/7202ed7206ebd13682c72821b88b0f36b68ff36d/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpCycleVariant.kt#L25)) is 0.

### LINKED ISSUE #1001: Power Distillation description is incorrect when two vectors are passed
### Modloader

Forge

### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Modloader version

Forge 47.4.10

### Modpack info

_No response_

### The latest.log file

_No response_

### Issue description

The hex notebook states under the description of Power Distillation: "With two vectors, combines them into the vector projection of the top of the stack onto the second-from-the-top." However, when testing, it seems to be swapped as listed here, so the second-from-the-top is actually projected onto the top.

Either the description or the functionality should be changed so that they match. In my opinion, changing the description would be best, as how it works seems more natural than the description's way, and it would prevent pre-existing hexes and tools from breaking.

### Steps to reproduce

_No response_

### Other information

I have MoreIotas 0.1.1 and Hexal 0.3.1 installed as well

Robotgiggle:
- Fixes the Thoth's Gambit description incorrectly implying that it uses a single pattern rather than a list of them
- Changes the listed size limit of an Akashic Library to the correct value (128 blocks from the record)
- Removes the mention of a range limit from the Greater Teleport description, since the limit was removed in 1.20
- Trims down the descriptions for Hermes' Gambit and Greater Teleport (while preserving the key information) to avoid text overflowing the page
- Fixes #977 by adding a lang key for `a placeable item` along with a new mishap for incorrect hotbar items
- Fixes #989 by setting the cost of Caster's Glamour to 0.1 dust and updating the documentation to match
- Fixes #1001 by updating the Power Distillation description to match its actual behavior with vectors


Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapBadHotbarItem.kt
```diff
@@ -0,0 +1,31 @@
...
+import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
+import at.petrak.hexcasting.api.casting.iota.Iota
+import at.petrak.hexcasting.api.pigment.FrozenPigment
+import at.petrak.hexcasting.api.utils.asTranslatedComponent
+import net.minecraft.network.chat.Component
+import net.minecraft.world.InteractionHand
+import net.minecraft.world.item.DyeColor
+import net.minecraft.world.item.ItemStack
+
+class MishapBadHotbarItem(val item: ItemStack?, val wanted: Component) : Mishap() {
```
object-Object: Nit: can we add a doc comment to this class?

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -2105,8 +2108,8 @@
...
+        "1": "Far more powerful than $(l:patterns/spells/basic#hexcasting:blink)$(action)Blink/$, this spell lets me teleport an entity to anywhere in the entire world! No matter the distance, it always seems to cost about ten $(l:items/amethyst)$(item)Charged Amethyst/$.",
+        "2": "The provided vector is an offset, not an absolute position in the world; for example, if I use (1, 0, 0), the entity will end up precisely one block east of its original position.$(br2)Unfortunately, it seems that when teleporting myself, I'm unable to maintain a perfect focus. This can cause some of my items to splatter out of my inventory on arrival.$(br2)I should also note that the target will be forcibly removed from anything inanimate they are riding ... but I've read scraps that suggest animals can come along for the ride, so to speak.",
```
object-Object: Nit: should we use a different word than "focus" (e.g. "maintain perfect concentration") to avoid confusion with the item?

object-Object (approved):
LGTM with two small nits; however, can this PR be given a more descriptive title before merging, since the title now determines the squashed commit message?

PR #1012, opened 2026-03-09
Userdata components

miyucomics:
Makes the `userData` NBT compound tag inside `CastingImage`s into a typed component system. Components are deserialized once along with the rest of the image—referring to its component type to determine how to deserialize—and they are serialized once along with the rest of the image.

This would prevent a large deal of unneeded serialization and deserialization for the ravenmind and features in addons that may take a long time to serialize and deserialize by only doing them once and working on a copy in memory. There has been reported massive slowdowns with the ravenmind in hexes that need to iterate often, and from my testing, this seems to have fixed it.

Also adds the concept of a transient component, which disappears when the 'cast' it is created in—from start to when the continuation queue empties at the end of `queueAndExecuteIotas` right before the casting environment `postCast` callback—ends. This fixes unintuitive behavior such as the +1 tax for impulses on the single entity persisting for the entire duration of a spell circle or staff cast, which can stick around arbitrarily long. Now, the jump between slates and right after a pattern is drawn, transient components are cleared.


Robotgiggle (changes requested):
This looks good, but it should be retarged to the 1.21 branch - removing `CastingImage.userData` is a breaking change.

PR #1021, opened 2026-03-31
Add config toggle for the iota NBT dump with advanced tooltips

Robotgiggle:
Currently, if Advanced Tooltips is enabled, iota-holding items will display the full NBT of their stored iota in their tooltip, which can get very unwieldy. This PR adds a client config toggle to enable/disable this feature. Since it's largely a debug tool, the config toggle is disabled by default.

The `HexConfig.client() == null` before the check for the actual config value is there in case `appendHoverText()` ever gets called on the server side on Fabric (which doesn't have a client config on the server). I have no idea if this is actually possible in vanilla+hex, but I figured better safe than sorry.


object-Object (approved):
LGTM

PR #1023, opened 2026-04-02
Several tweaks for brainsweep sacrifice entity display

YukkuriC:
- loop display every choices for tag ingredient
- cache them on client to prevent by-frame new instances


Robotgiggle (changes requested):
The cycling entity display only works in recipe viewers, not in the guidebook. Making it work in the guidebook is totally possible though, you'd just need to change some things around in [the component processor](<https://github.com/FallingColors/HexMod/blob/main/Common/src/main/java/at/petrak/hexcasting/interop/patchouli/BrainsweepProcessor.java>) for brainsweep recipes. That should probably get done before this gets merged.

s5bug:
@YukkuriC If you haven't started working on this change, don't worry about it, I'll implement it. If you have, let me know so we don't duplicate work!

Robotgiggle (approved):
Turns out Patchouli is in fact not capable of showing cycling entity displays, so we won't be able to implement that yet in this PR. 

However, some preliminary refactoring to turn `exampleEntity` into `exampleEntities` for ease of use has been done anyway, and an issue has been created in the Patchouli repo (VazkiiMods/Patchouli#852) to hopefully enable cycling entities at some point in the future.

PR #1025, opened 2026-04-04
Add handling for incorrect entity lookdirs

Robotgiggle:
Vanilla code has sign errors in the look direction vectors returned by [projectiles](https://bugs.mojang.com/browse/MC/issues/MC-112474) and [phantoms](https://bugs.mojang.com/browse/MC/issues/MC-134707). This PR adds special handling to account for that, and makes use of it in OpEntityLook and OpBlink.

I realize that a registration system like the one currently used for special velocity getters might be better in concept, but since this bug applies to literally all projectiles regardless of individual entity type, I figured it would be simpler to just do what I've done here.


object-Object (approved):
LGTM

PR #1027, opened 2026-04-12
fix: packaged items should batch pattern packets like staff casting does

### LINKED ISSUE #1026: Is this necessary to send the same hex twice for packaged hex item casting
https://github.com/FallingColors/HexMod/blob/532fe9a60138544112e096812c7aefb78b3d7364/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PackagedItemCastEnv.java#L28
and
https://github.com/FallingColors/HexMod/blob/532fe9a60138544112e096812c7aefb78b3d7364/Common/src/main/java/at/petrak/hexcasting/common/items/magic/ItemPackagedHex.java#L138

considering there's cooldown and removing-dupe logic client-side, this could be a waste

pythonmcpi:
Previous behavior:
Casting with a packaged item results in the server sending one packet per evaluated pattern, causing network lag

Fixed behavior:
Casting with a packaged item will batch the patterns in a set and send it all in one packet at the end, like how the StaffCastEnv already does.

Closes #1026

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
@@ -22,6 +22,9 @@ public abstract class PlayerBasedSpiralPatternCastEnv extends PlayerBasedCastEnv
         super(caster, castingHand);
     }
 
+    /**
+     * Returns the duration (in ticks) that patterns should remain visible after being executed by the env.
+     */
     public abstract int getSpiralPatternDuration();
 
     @Override
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
@@ -1,0 +1,48 @@
+package at.petrak.hexcasting.api.casting.eval.env;
+
+import at.petrak.hexcasting.api.casting.eval.CastResult;
+import at.petrak.hexcasting.api.casting.eval.vm.CastingImage;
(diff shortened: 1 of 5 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Variant of {@link PlayerBasedCastEnv} that tracks executed patterns and displays them in a spiral around the player
+ */
+public abstract class PlayerBasedSpiralPatternCastEnv extends PlayerBasedCastEnv {
+    private final Set<HexPattern> castPatterns = new HashSet<>();
+
+    public PlayerBasedSpiralPatternCastEnv(ServerPlayer caster, InteractionHand castingHand) {
+        super(caster, castingHand);
+    }
```
object-Object, line 23: Nit: does this constructor need to exist if it doesn't do anything?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public abstract int getSpiralPatternDuration();
```
object-Object, line 28: Can you add a comment explaining what this is for and what unit it's in?

object-Object (commented):
Mostly LGTM! Just a possible cleanup and a place where a comment would be useful

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/env/PlayerBasedSpiralPatternCastEnv.java
pythonmcpi: It's required by java, removing it breaks building.

object-Object (approved):

PR #1019, opened 2026-03-27
Fix `invertAbacusScrollDirection` using wrong default value on fabric

aloaloolaola:
fixes #1018

Issue #1018, opened 2026-03-27
Abacus scroll setting uses spellbook setting's default value


aloaloolaola:
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

On fabric, the invertAbacusScrollDirection config uses DEFAULT_INVERT_SPELLBOOK_SCROLL instead of DEFAULT_INVERT_ABACUS_SCROLL [here](https://github.com/FallingColors/HexMod/blob/532fe9a60138544112e096812c7aefb78b3d7364/Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexConfig.java#L134).

Robotgiggle:
Fixed in #1028

PR #1028, opened 2026-04-12
Add keybinds for spellbook control and config to disable in-world scrolling

### LINKED ISSUE #373: Accessibility request: alternative to scroll wheel for changing the value of a spell book, abacus, etc.
as I explained on discord, scroll wheel controls can be difficult for people with disabilities such as my own. In my case, I have a spinal cord injury, and both standard gaming mice and physical keyboards are... difficult... for me to interface with. I have a trackball mouse, and use virtual alternatives for keyboard functions. Neither of these options provide a physical or virtual scroll wheel, so I would have to further work around it in something like AutoHotkey (to rebind the -/= keys, for example, to the mouse wheel as a virtual hotkey) or pull in another input source (but I would still have to use AutoHotkey to forward the joystick control I have).
In any case, could you please make it possible to rebind the mouse wheel controls to something else? I have a few possible controls in mind (mentioned in the parentheses above) that I would like to use, but I presume, given I don't see it in the Keybinds menu, that the current mouse wheel control is hardcoded so I can't use either of them.

pythonmcpi:
This PR includes two changes:
- Keybinds (unbound by default) for switching spellbook pages (closes #373)
- Clientside config setting to disable in-world scrolling of spellbooks and abaci
  - Changes supersede #1018 and #1019, which I did not know existed until moments ago
  - Fixes invertAbacusScrollDirection's fabric tooltip to describe changing the "value" rather than "page index"

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/mod/HexConfig.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/mod/HexConfig.java
@@ -41,6 +41,8 @@ public class HexConfig {
     public interface ClientConfigAccess {
         boolean ctrlTogglesOffStrokeOrder();
 
+        boolean disableInworldScrolling();
+
         boolean invertSpellbookScrollDirection();
 
         boolean invertAbacusScrollDirection();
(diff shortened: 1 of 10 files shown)


Robotgiggle:
Is it intended that enabling the "invert spellbook scroll direction" config also inverts the functionality of the new keybinds? 

The keybinds call `ShiftScrollListener.onScroll()` and the spellbook inversion config is checked _within_ `ShiftScrollListener` after that method has been processed.

Twisted-Code:
As your enlightened neighborhood quadriplegic, I appreciate more configurable control mappings <3

PR #1031, opened 2026-04-28
TreeList for asymptotically-efficient immutable lists [part 1: Thoth's]

### LINKED ISSUE #804: Iris with Thoths crash
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.2-pre-701

### Modloader version

Fabric Loader 0.16.9 + API 0.92.2

### Modpack info

_No response_

### The latest.log file

https://mclo.gs/ey0FmkO

### Issue description

Crashes when running Iris inside Thoths

the most likely issue is using Iris with Thoths can cause thier references to loop and a serialization recursion function goes in loops indefinitely and eventually causes stack overflow, crashing the server

### Steps to reproduce

1. Cast the following
Introspection
Introspection
Iris Gambit
Retrospection
Vacant Reflection
Single's Purification
Thoth's Gambit
Retrospection
Hermes' Gambit
2. crash

### Other information

sorry for more work

s5bug:
Fixes #804
Conflicts with #897 

This is part 1, part 2 will change ListIota to use TreeList, part 3 will change the VM to use TreeList for the stack.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
@@ -6,6 +6,7 @@ import at.petrak.hexcasting.api.casting.eval.ResolvedPatternType
 import at.petrak.hexcasting.api.casting.iota.Iota
 import at.petrak.hexcasting.api.casting.iota.ListIota
 import at.petrak.hexcasting.api.utils.NBTBuilder
+import at.petrak.hexcasting.api.utils.TreeList
 import at.petrak.hexcasting.api.utils.getList
 import at.petrak.hexcasting.api.utils.hasList
 import at.petrak.hexcasting.api.utils.serializeToNBT
@@ -28,14 +29,13 @@ data class FrameForEach(
     val data: SpellList,
     val code: SpellList,
     val baseStack: List<Iota>?,
-    val acc: MutableList<Iota>
+    val acc: TreeList<Iota>
 ) : ContinuationFrame {
 
     /** When halting, we add the stack state at halt to the stack accumulator, then return the original pre-Thoth stack, plus the accumulator. */
     override fun breakDownwards(stack: List<Iota>): Pair<Boolean, List<Iota>> {
         val newStack = baseStack?.toMutableList() ?: mutableListOf()
-        acc.addAll(stack)
-        newStack.add(ListIota(acc))
+        newStack.add(ListIota(acc.appendedAll(stack).toList()))
         return true to newStack
     }
 
@@ -46,13 +46,12 @@ data class FrameForEach(
         harness: CastingVM
     ): CastResult {
         // If this isn't the very first Thoth step (i.e. no Thoth computations run yet)...
-        val stack = if (baseStack == null) {
-            // init stack to the VM stack...
-            harness.image.stack.toList()
+        val (stack, nextAcc) = if (baseStack == null) {
+            // init stack to the harness stack...
+            harness.image.stack.toList() to acc
         } else {
             // else save the stack to the accumulator and reuse the saved base stack.
-            acc.addAll(harness.image.stack)
-            baseStack
+            baseStack to acc.appendedAll(harness.image.stack)
         }
 
         // If we still have data to process...
@@ -60,13 +59,13 @@ data class FrameForEach(
             // push the next datum to the top of the stack,
             val cont2 = continuation
                 // put the next Thoth object back on the stack for the next Thoth cycle,
-                .pushFrame(FrameForEach(data.cdr, code, stack, acc))
+                .pushFrame(FrameForEach(data.cdr, code, stack, nextAcc))
                 // and prep the Thoth'd code block for evaluation.
                 .pushFrame(FrameEvaluate(code, true))
             Triple(data.car, harness.image.withUsedOp(), cont2)
         } else {
             // Else, dump our final list onto the stack.
-            Triple(ListIota(acc), harness.image, continuation)
+            Triple(ListIota(acc.toList()), harness.image, continuation)
         }
         val tStack = stack.toMutableList()
         tStack.add(stackTop)
@@ -86,7 +85,7 @@ data class FrameForEach(
         "code" %= code.serializeToNBT()
         if (baseStack != null)
             "base" %= baseStack.serializeToNBT()
-        "accumulator" %= acc.serializeToNBT()
+        "accumulator" %= acc.toList().serializeToNBT()
     }
 
     override fun size() = data.size() + code.size() + acc.size + (baseStack?.size ?: 0)
@@ -104,10 +103,10 @@ data class FrameForEach(
                         HexIotaTypes.LIST.deserialize(tag.getList("base", Tag.TAG_COMPOUND), world)!!.list.toList()
                     else
                         null,
-                    HexIotaTypes.LIST.deserialize(
+                    TreeList.from(HexIotaTypes.LIST.deserialize(
                         tag.getList("accumulator", Tag.TAG_COMPOUND),
                         world
-                    )!!.list.toMutableList()
+                    )!!.list)
                 )
             }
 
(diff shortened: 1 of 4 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
```diff
@@ -28,14 +29,13 @@ data class FrameForEach(
     val data: SpellList,
     val code: SpellList,
     val baseStack: List<Iota>?,
-    val acc: MutableList<Iota>
+    val acc: TreeList<Iota>
```
object-Object: It'd be nice if we could try to avoid breaking existing addons that construct a `FrameForEach` or access `acc`. To that end, could we do the following:
- Rename this `acc` to something else, maybe `immutableAcc` or `treeAcc` or something
- Add a deprecated constructor that takes `List<Iota>` and converts it to a `TreeList`
- Add a deprecated `val acc: MutableList<Iota> get() = immutableAcc.toMutableList()` so existing code that reads `acc` doesn't crash

This'll still break any addons that use `.copy(acc=)`, but I didn't see any instances of that in a GitHub search.

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
```diff
@@ -28,14 +29,13 @@ data class FrameForEach(
...
 ) : ContinuationFrame {
 
     /** When halting, we add the stack state at halt to the stack accumulator, then return the original pre-Thoth stack, plus the accumulator. */
     override fun breakDownwards(stack: List<Iota>): Pair<Boolean, List<Iota>> {
         val newStack = baseStack?.toMutableList() ?: mutableListOf()
-        acc.addAll(stack)
-        newStack.add(ListIota(acc))
+        newStack.add(ListIota(acc.appendedAll(stack).toList()))
```
object-Object: Would it be worth adding a `ListIota` constructor that takes a `TreeList<Iota>`? Then, when #1032 changes `ListIota` to store a `TreeList<Iota>` internally, existing cases like this (or in addons) would automatically use the more efficient path instead of unnecessarily converting to a list and back.

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
```diff
@@ -46,27 +46,26 @@ data class FrameForEach(
         harness: CastingVM
     ): CastResult {
         // If this isn't the very first Thoth step (i.e. no Thoth computations run yet)...
-        val stack = if (baseStack == null) {
-            // init stack to the VM stack...
-            harness.image.stack.toList()
+        val (stack, nextAcc) = if (baseStack == null) {
```
object-Object: Nit: could this be called `newAcc` to match other similar variable names around here?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
```diff
@@ -86,7 +85,7 @@ data class FrameForEach(
         "code" %= code.serializeToNBT()
         if (baseStack != null)
             "base" %= baseStack.serializeToNBT()
-        "accumulator" %= acc.serializeToNBT()
+        "accumulator" %= acc.toList().serializeToNBT()
```
object-Object: Nit: is this `.toList()` needed? One of the `serializeToNBT()` extension functions is for `Iterable<Iota>`, which I'd expect `TreeList<Iota>` to count as

https://github.com/FallingColors/HexMod/blob/724c36bba6a97f97d16f95d16f7addb700e62443/Common/src/main/java/at/petrak/hexcasting/api/utils/HexUtils.kt#L264

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
```diff
@@ -46,27 +46,26 @@ data class FrameForEach(
...
                 // put the next Thoth object back on the stack for the next Thoth cycle,
-                .pushFrame(FrameForEach(data.cdr, code, stack, acc))
+                .pushFrame(FrameForEach(data.cdr, code, stack, nextAcc))
                 // and prep the Thoth'd code block for evaluation.
                 .pushFrame(FrameEvaluate(code, true))
             Triple(data.car, harness.image.withUsedOp(), cont2)
         } else {
             // Else, dump our final list onto the stack.
-            Triple(ListIota(acc), harness.image, continuation)
+            Triple(ListIota(acc.toList()), harness.image, continuation)
```
object-Object: Another case that could be simplified with a `ListIota(TreeList<Iota>)` constructor

Common/src/main/java/at/petrak/hexcasting/api/utils/TreeList.java
```diff
@@ -0,0 +1,3386 @@
...
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import static java.util.Arrays.copyOf;
+import static java.util.Arrays.copyOfRange;
+
+// TODO mark everything with use-site variance
+
+/**
+ * Ported from Scala 2.13.16's scala.collection.immutable.Vector
```
object-Object, line 15: Could this also describe what the class is for, not just where it's ported from?

Common/src/main/java/at/petrak/hexcasting/api/utils/TreeList.java
```diff
@@ -0,0 +1,3386 @@
...
+import java.lang.reflect.Array;
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import static java.util.Arrays.copyOf;
+import static java.util.Arrays.copyOfRange;
+
+// TODO mark everything with use-site variance
```
object-Object, line 12: What's this mean?

Common/src/main/java/at/petrak/hexcasting/api/utils/TreeList.java
```diff
@@ -0,0 +1,3386 @@
...
+    public TreeList<A> reversed() {
+        // FIXME: Understand NewVectorIterator
+        VectorBuilder<A> builder = new VectorBuilder<>();
+        for(int i = this.size() - 1; i >= 0; i--) {
+            builder.addOne(this.get(i));
+        }
+        return builder.result();
+    }
+
+    // FIXME: Get rid of this
```
object-Object: What are these FIXME comments for, and/or when will they be fixed?

Common/src/main/java/at/petrak/hexcasting/api/utils/TreeList.java
object-Object, line 1: For transparency - I mostly just skimmed this file, don't have the energy to review 3300 lines of Java right now. General notes:
- Would it be worth bulk replacing the various instances of `vector`/`Vector` with `treeList`/`TreeList` respectively, or at least the public ones?
- Can the various public members be moved to the top of their classes so it's easier for consumers to see what methods are available?
- Could maybe use more doc comments on public members
- Very optional, but some unit tests might be nice if it's not too much effort

object-Object (commented):
Mostly LGTM! Comments are mostly cleanups for maintainability, and suggestions to reduce addon breakage

Common/src/main/java/at/petrak/hexcasting/api/utils/TreeList.java
s5bug: There's many places where `Iterable<A>` is used when `Iterable<? extends A>` should be used instead. It hasn't shown to be a problem in real usage, so I haven't bothered to fix it yet.

Common/src/main/java/at/petrak/hexcasting/api/utils/TreeList.java
s5bug: I believe there's an issue where Kotlin tries to assume `reversed` has a type of `List<A>` rather than `TreeList<A>`.

The reason the `NewVectorIterator` comment is there is because I suspect a performance operation / being able to skip out on the cost faced by not "remembering" where one is in a sub-array. (`NewVectorIterator`'s forward iteration is really efficient and dodges a lot of the cost that indexing in a loop would face.)

PR #1032, opened 2026-04-28
Replace SpellList with TreeList in ListIota

s5bug:

PR #1033, opened 2026-04-28
Replace mutable list usages with TreeList in stack operations

s5bug:
Stacked on #1032 

TODO is make a pass through _all_ the list ops and make sure they're not doing O(n) copies but I'm not going to waste my time with that in case I have to go back and modify other stuff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - Removed APIs deprecated in `0.11.4` ([#1126](https://github.com/FallingColors/HexMod/pull/1126), [#1127](https://github.com/FallingColors/HexMod/pull/1127) [#1129](https://github.com/FallingColors/HexMod/pull/1129), [#1137](https://github.com/FallingColors/HexMod/pull/1137), [#1142](https://github.com/FallingColors/HexMod/pull/1142)) @s5bug
 - ListIota now uses the asymptotically-efficient TreeList internally ([#1032](https://github.com/FallingColors/HexMod/pull/1032)) @s5bug
-- SpellList has been removed, the Casting Image and Casting Frames now store iotas in a TreeList ([#1033](https://github.com/FallingColors/HexMod/pull/1033)) @s5bug
+- SpellList has been removed ([#1032](https://github.com/FallingColors/HexMod/pull/1032)) @s5bug
+- Casting Image and Casting Frames now store iotas in a TreeList ([#1033](https://github.com/FallingColors/HexMod/pull/1033)) @s5bug
 
 ## `0.11.3` - 2025-11-22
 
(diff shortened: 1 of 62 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - Removed APIs deprecated in `0.11.4` ([#1126](https://github.com/FallingColors/HexMod/pull/1126), [#1127](https://github.com/FallingColors/HexMod/pull/1127) [#1129](https://github.com/FallingColors/HexMod/pull/1129), [#1137](https://github.com/FallingColors/HexMod/pull/1137), [#1142](https://github.com/FallingColors/HexMod/pull/1142)) @s5bug
 - ListIota now uses the asymptotically-efficient TreeList internally ([#1032](https://github.com/FallingColors/HexMod/pull/1032)) @s5bug
+- SpellList has been removed, the Casting Image and Casting Frames now store iotas in a TreeList ([#1033](https://github.com/FallingColors/HexMod/pull/1033)) @s5bug
```
Robotgiggle: SpellList was removed in #1032, not here

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/stack/OpAlwinfyHasAscendedToABeingOfPureMath.kt
```diff
@@ -42,7 +43,9 @@ object OpAlwinfyHasAscendedToABeingOfPureMath : Action {
             editTarget = editTarget.subList(1, editTarget.size)
         }
 
-        val image2 = image.withUsedOp().copy(stack = stack)
+        // FIXME this is expensive if the stack is incredibly large
+        // someone should rewrite this op to cut and replace a `slice` of the `TreeList` stack
```
Robotgiggle: I think I have an implementation for this - can I push it to this PR or do you want it in a separate PR after this goes in?

Robotgiggle (commented):
I haven't tested exhaustively but nothing seems to be broken. Simulate still works fine, as does the Shepherd Directrix.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/stack/OpAlwinfyHasAscendedToABeingOfPureMath.kt
s5bug: Go nuts 👍

Robotgiggle: Ok, it's been pushed. Let me know if you think I did anything wrong.

Robotgiggle: Is there any advantage to having `newOrder` be a `TreeList`? Or would it be better to have it also be a standard `MutableList`?

s5bug:
New Swindler's looks good, I just replaced the `x = x.addOne` with `x` being a builder instead since the collection isn't used for anything else

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/stack/OpAlwinfyHasAscendedToABeingOfPureMath.kt
s5bug: At such a small size I think it makes almost no difference. If we know the "eventual" size of `newOrder` then I'd say make it an `ArrayList` pre-allocated to that size.

s5bug: There's an asymptotic fast-path for if the thing being appended is already a treelist:
- With an arbitrary non-TreeList collection, `appendedAll` looks like `newBuilder.append(this).append(that).build`, O(m+n)
- If either `this` or `that` are relatively small enough, the fast path is O(min(m, n))

So, I'd say maybe just leave it a TreeList if we're optimizing for the "people are lagging the server" case, but I don't know 😄 not experienced in that area

Robotgiggle (approved):
LGTM

PR #1038, opened 2026-05-10
Replace toList and mutableList usages with TreeList in all operations

s5bug:
Stacked on #1033


Common/src/main/java/at/petrak/hexcasting/common/casting/arithmetic/operator/list/OperatorUnappend.kt
```diff
@@ -13,8 +13,8 @@ import at.petrak.hexcasting.common.lib.hex.HexIotaTypes.LIST
 object OperatorUnappend : OperatorBasic(1, IotaMultiPredicate.all(IotaPredicate.ofType(LIST.get()))) {
     override fun apply(iotas: Iterable<Iota>, env: CastingEnvironment): Iterable<Iota> {
         val it = iotas.iterator().withIndex()
-        val list = it.nextList(arity).toMutableList()
-        val last = list.removeLastOrNull() ?: NullIota()
-        return listOf(ListIota(list), last)
+        val list = it.nextList(arity)
+        val last = if(list.isEmpty()) NullIota() else list.last()
+        return listOf(ListIota(list.init()), last)
```
Robotgiggle: If the list was empty, this attempts to call `init()` on an empty list, which causes an UnsupportedOperationException

Robotgiggle (changes requested):
Everything looks good except the problem noted below

Robotgiggle (approved):
LGTM

PR #1040, opened 2026-05-10
Wayfarer/anchorite cost rebalance

Robotgiggle:
- Reduces cost of Wayfarer's Flight to 2 dust per second
- Increases cost of Anchorite's Flight to 2 dust per meter of radius
- Anchorite's Flight now has a minimum cost of 2 dust

Anchorite's Flight has always been remarkably cheap for what it does, and Wayfarer's Flight is so much more expensive that there's hardly any situation where you'd want to use it. This aims to fix both of those problems.

Also, adding a minimum cost to Anchorite prevents people from using it to go straight up and down (or block fall damage) effectively for free by providing an extremely small fractional radius.


poolcritter:
Perhaps we make the minimum of 1 metre only apply to the _cost_, as to avoid breaking existing Hexes?

Robotgiggle:
Hmm, good point

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFlight.kt
```diff
@@ -35,11 +35,10 @@ class OpFlight(val type: Type) : SpellAction {
...
 
-        val cost = when (this.type) {
-            Type.LimitRange -> theArg * MediaConstants.DUST_UNIT
-            // A second of flight should cost 1 shard
-            Type.LimitTime -> theArg * MediaConstants.SHARD_UNIT
-        }.roundToLong()
+        // One block of radius, or one second of duration, costs 2 dust
+        var cost = (theArg * 2 * MediaConstants.DUST_UNIT).roundToLong()
+        // Cost for anchorite does not decrease below 1 meter
+        if (type == Type.LimitRange) cost = max(cost, 2 * MediaConstants.DUST_UNIT)
```
object-Object: Nit: can we refactor the `2 * MediaConstants.DUST_UNIT` into a separate variable so the logic is a bit more clear?

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Changed
 
 - Changed the `media_consumption` attribute to only apply to player-based casting, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987).
+- Decreased the cost of Wayfarer's Flight to 2 dust per second, by Robotgiggle in [#1040](https://github.com/FallingColors/HexMod/pull/1040).
+- Increased the cost of Anchorite's Flight to 2 dust per meter and capped the cost at 2 dust for radii below 1 meter, by Robotgiggle in [#1040](https://github.com/FallingColors/HexMod/pull/1040).
```
object-Object: Nit: could probably just put both of these in one line with a bit less detail

object-Object (approved):
LGTM with a couple of nits

PR #1041, opened 2026-05-11
Config options for rescaling media costs

Robotgiggle:
Adds a list in the config that can be used to rescale the cost of any action. Entries consist of a resource location and a number (ie `hexcasting:explode 2.5`) and the cost of the listed action is scaled by the listed number.

Also adds a config value with a global scaling factor that will apply to _all_ media costs, if you just want to apply an overall buff or nerf to hex in your pack/server.

Both of these scaling factors apply before the `media_consumption_modifier` attribute. The global scaling factor is not applied to anything in the `cannot_modify_cost` tag, so that making everything cheaper doesn't immediately introduce a bunch of exploits. However, entries in the per-action list for actions in that tag _do_ still work, in case for whatever reason you actually do want to rescale one of those patterns.


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/CastingEnvironment.java
IridescentVoid: Nit: Might be worth noting that this also modifies some per-pattern CastingEnv state, rather than (otherwise pure) checks.

Forge/src/main/java/at/petrak/hexcasting/forge/ForgeHexConfig.java
```diff
@@ -185,10 +187,30 @@ public Server(ForgeConfigSpec.Builder builder) {
...
                 "0 = wood, 1 = stone, 2 = iron, 3 = diamond, 4 = netherite."
             ).defineInRange("opBreakHarvestLevel", DEFAULT_OP_BREAK_HARVEST_LEVEL, 0, 4);
+
+            greaterTeleportSplatsItems = builder.comment(
+                    "Should items fly out of the player's inventory when using Greater Teleport?"
+            ).define("greaterTeleportSplatsItems", DEFAULT_GREATER_TELEPORT_SPLATS_ITEMS);
+
+            actionDenyList = builder.comment(
+                    "Resource locations of disallowed actions. Trying to cast one of these will result in a mishap. " +
+                        "For example, \"hexcasting:get_caster\" will prevent Mind's Reflection.")
```
IridescentVoid: The quote usage here around the action resourcelocations diverge from the language strings used for the fabric config.

Forge/src/main/java/at/petrak/hexcasting/forge/ForgeHexConfig.java
```diff
@@ -185,10 +187,30 @@ public Server(ForgeConfigSpec.Builder builder) {
...
+
+            costRescaleList = builder.comment(
+                    "Maps resource locations to the scaling factor for that specific action's media cost. " +
+                        "For example, \"hexcasting:add_motion 3\" will make Impulse cost 3x as much.")
+                .defineList("costRescaleList",  List.of(), Server::isValidReslocDoublePair);
+
+            globalCostScaling = builder.comment(
+                    "All media costs, except for actions in the #hexcasting:cannot_modify_cost tag, will be multiplied by this value." +
+                        "If you want to modify the cost of an action in that tag, use the per-action list above.")
+                .defineInRange("globalCostScaling", DEFAULT_GLOBAL_COST_SCALING, 0.0, Double.MAX_VALUE);
```
IridescentVoid, line 217: Nit: some of the forge config range validation (including ones not touched in this pr, such as opBreakHarvestLevel) seems to be missing on fabric.

Forge/src/main/java/at/petrak/hexcasting/forge/ForgeHexConfig.java
```diff
@@ -254,6 +267,20 @@ public boolean isActionAllowedInCircles(ResourceLocation actionID) {
             return noneMatch(circleActionDenyList.get(), actionID);
         }
 
+        @Override
+        public double getActionCostScaling(ResourceLocation actionID) {
```
IridescentVoid, line 275: Could we cache this instead of re-parsing on every pattern?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 - Added the `cannot_modify_cost` tag for patterns that should ignore the `media_consumption` attribute when calculating cost, by Robotgiggle in [987](https://github.com/FallingColors/HexMod/pull/987).
+- Added a config options to rescale the cost of some or all patterns ([#1041](https://github.com/FallingColors/HexMod/pull/1041)) @Robotgiggle
```
IridescentVoid: Small typo: "a config options"

Nit: Wording could be improved for clarity.
```suggestion
- Added config options to scale pattern costs per action and globally ([#1041](https://github.com/FallingColors/HexMod/pull/1041)) @Robotgiggle
```

IridescentVoid (approved):
Looks good to merge once comments are addressed.

PR #1043, opened 2026-05-11
Update Gradle, Kotlin, and Loom to support Inline 1.2.2

Robotgiggle:
The current implementation of the pattern overlay feature added in #879 relies on a method in Inline that was renamed in the 1.2.0 update. Thus, if anyone uses a version of Inline beyond that point with the current version of hex, it will break. This PR updates hex's Inline dependency to the latest version and updates the relevant methods to their new names to prevent that from happening, and also updates a bunch of core libraries which are required for the new Inline version.

Note that if this PR does _not_ get merged before the 0.11.4 release, we should instead change the Inline dependency in `fabric.mod.json` (fabric) and `mods.toml` (forge) to explicitly exclude Inline versions 1.2.0 and onwards.

Also note that the new Kotlin version (2.2.21) now gives a bunch of warnings about the use of public `copy()` methods on data classes with private constructors, since that functionality will be removed in future versions of Kotlin. More info can be found [here](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-consistent-copy-visibility/) and [here](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-exposed-copy-visibility/).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -60,7 +60,7 @@ jobs:
             mc-runtime-test: fabric
             fabric-api: 0.87.0
             dependencies: >-
-              'https://cdn.modrinth.com/data/Ha28R6CL/versions/vlhvI5Li/fabric-language-kotlin-1.10.18%2Bkotlin.1.9.22.jar'
+              'https://cdn.modrinth.com/data/Ha28R6CL/versions/LcgnDDmT/fabric-language-kotlin-1.13.7%2Bkotlin.2.2.21.jar'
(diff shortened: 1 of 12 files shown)


.github/workflows/pr.yml
object-Object, line 1: Do we also need to update the versions for fabric-language-kotlin and kotlinforforge?

Fabric/src/main/resources/fabric.mod.json
object-Object, line 1: Same here and in mods.toml - should we increase the minimum required kotlin versions?

object-Object (approved):
LGTM, but wondering if we should also bump the versions of the kotlin library mods.

Robotgiggle:
Hmm yeah, probably should update those too

PR #1046, opened 2026-05-13
Advanced parenthesis patterns

Robotgiggle:
**NOTE:** Due to the changes from #1047, this PR needed a substantial rewrite, so I closed it and made a new PR (#1103).

Adds Meditation and Recollection, advanced versions of the standard intro/retro patterns. 
- Meditation pops an integer and opens a list to the specified parenthesis depth.
- Recollection closes a list from any parenthesis depth, pushing the depth and the list.

Meditation is treated as a normal pattern when inside existing parentheses because there's no way to determine how many new layers it should open. This is kind of a janky solution given that every other pattern in this section of the book works even while nested, but the only alternative would be to prevent using it inside parens altogether.

Recollection pushes the list on top of the depth for convenience, since you're much more likely to need the list as an argument for something (saving it to a focus, adding something with Integration Distillation, etc) than the number.

Screenshot 2026-05-13 000048


Real-Luxof:
Can they be named something cooler? Like Continue and Cessation/Cease?

Robotgiggle:
I'm definitely down to change the names to something better, but I feel like Resumption/Cessation (the names from Hexic) don't really have quite the same vibe as Introspection/Retrospection. Also, we probably shouldn't use Continuation to avoid confusion with jumps and the continuation frame system.

Real-Luxof:
Could be Meditation and Afterthought maybe

Real-Luxof:
There's no page for the nested big introspection mishap.

Robotgiggle:
> There's no page for the nested big introspection mishap.

Ah, good catch. I'll get that done shortly.

Robotgiggle:
Technically yes? But I think it's a little better from an organizational perspective to keep everything together, especially since making it an `Action` would require making a whole new subfolder in the actions folder or just leaving it awkwardly in the top level.

If more people think it should be an `Action` I can move it over, I just think it makes a little more sense this way.

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -123,8 +124,12 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
                         OperatorSideEffect.DoMishap(
                             e,
                             Mishap.Context(
-                                (iota as? PatternIota)?.pattern ?: HexPattern(HexDir.WEST),
-                                HexAPI.instance().getRawHookI18n(HexAPI.modLoc("close_paren"))
+                                pattern ?: HexPattern(HexDir.WEST),
+                                HexAPI.instance().getRawHookI18n(HexAPI.modLoc(when (pattern?.angles) {
+                                    SpecialPatterns.RETROSPECTION.angles -> "close_paren"
+                                    SpecialPatterns.MEDITATION.angles -> "open_n_parens"
```
Real-Luxof, line 130: Meditation never throws now, so I think this part should be removed (correct me if I'm wrong)

Robotgiggle: It still throws if you cast it without the proper number argument (see lines 310 and 311)

Real-Luxof: ic

Real-Luxof (approved):
- lgtm
- can't think of a reason why this shouldn't be in base hex

c-Caelum:
This is neat. I like this. Maybe, though, Meditation could add some given list to the intro/retro that's made?

Robotgiggle:
I did think of that (even went as far as implementing a basic version, actually) but then I realized that there's no way to properly figure out the correct paren depth using only an already-formed list iota, and if we ignore paren depth and just tack new stuff onto an existing list at depth 0, you can already do the same thing with additive dist. 

I realize it _would_ still be one pattern shorter, but I don't really think "pattern that just replicates a couple of existing patterns with no extra functionality" really has a place in base hex.

Robotgiggle:
Marked as draft since the whole thing will need to be reworked anyway once #1047 gets merged.

Robotgiggle:
Closing in favor of #1103.

PR #1047, opened 2026-05-14
Add proper support for escape manipulation by Actions

### LINKED ISSUE #1077: Evanition has a weird failure state if there's nothing to undo
### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

If Evanition is drawn when the in-progress parenthesized list does not contain anything (ie if the first thing you draw is Evanition, or if you draw enough Evanitions to undo all the patterns in the list) then the Evanition pattern will be displayed on the hex grid in red, like an invalid pattern, and no mishap or other indication of failure will happen.

This is hardcoded behavior in Evanition's implementation, not a side effect of failing to match its pattern or anything like that. With the upcoming rework to the pattern-escaping system which allows mishaps during parentheses-building, it would make much more sense to have a proper mishap here to indicate what went wrong.

Robotgiggle:
Adds the `operateInParens` method to allow `Action`s to define custom behavior when executed inside parentheses, and removes all the hardcoded jank that was used to do this before. 

This enables the previously hardcoded patterns for escape manipulation (Consideration, Introspection, Retrospection, and Evanition) to be turned into normal `Action`s, and also makes it possible to add new escape manipulators in addons or future updates.

Also fixes #1077 by allowing Evanition to undo the opening Introspection if there's nothing else to undo.


Common/src/main/java/at/petrak/hexcasting/api/casting/eval/SpecialPatterns.java
object-Object, line 1: Removing this file would be a breaking change - we should keep it and just deprecate it for now.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java
```diff
@@ -49,22 +49,34 @@ protected Iota(@NotNull IotaType<?> type, @NotNull Object payload) {
...
 
     /**
-     * This method is called when this iota is executed (i.e. Hermes is run on a list containing it, unescaped).
-     * By default it will return a {@link CastResult} indicating an error has occurred.
+     * This method is called when this iota is executed (i.e. Hermes is run on a list containing it).
+     * By default, if the iota is not within parentheses, it will return a {@link CastResult} indicating an error has occurred.
+     * If the iota *is* within parentheses, it will instead be added to the parenthesized list.
+     * Iotas can also be escaped using Consideration, but that's handled in {@link CastingVM#executeInner}.
      */
     public @NotNull CastResult execute(CastingVM vm, ServerLevel world, SpellContinuation continuation) {
```
object-Object, line 55: I think this is a breaking change - any addon that implements custom iota execution (and also ContinuationIota) will now execute their unescaped behaviour inside of parentheses until patched. Likely we should do the same thing as with Action (ie add an executeInParens method).

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapNeedsParens.kt
```diff
@@ -6,7 +6,7 @@ import at.petrak.hexcasting.api.casting.iota.PatternIota
 import at.petrak.hexcasting.api.pigment.FrozenPigment
 import net.minecraft.world.item.DyeColor
 
-class MishapTooManyCloseParens : Mishap() {
```
object-Object, line 9: Technically renaming this class is a breaking change, though I'm not sure any addons would be using it. Would be worth at least doing a GitHub search to check.

Common/src/main/java/at/petrak/hexcasting/api/HexAPI.java
object-Object, line 1: Removing public methods is a breaking change - we should instead deprecate them.

object-Object (changes requested):
Didn't fully review (am on a phone), but skimmed and noticed some problematic breaking changes.

Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapNeedsParens.kt
Robotgiggle: Based on GitHub search, the only addon that uses `MishapTooManyCloseParens` is HexThings - but HexThings also has a mixin to `CastingVM.handleParentheses`, so this PR breaks it even without the name change.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java
Robotgiggle: Good catch! This has now been fixed.

Common/src/main/java/at/petrak/hexcasting/api/HexAPI.java
```diff
@@ -62,9 +62,7 @@ default String getSpecialHandlerI18nKey(ResourceKey<SpecialHandler.Factory<?>> a
         return "hexcasting.special.%s".formatted(action.location().toString());
     }
 
-    /**
-     * Currently introspection/retrospection/consideration are hardcoded, but at least their names won't be
-     */
+    @Deprecated(since = "0.11.4")
```
object-Object, line 68: Nit: can we also add a javadoc `@deprecated` tag explaining why this is deprecated? Also applies to all other newly added `@Deprecated` annotations

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java
```diff
@@ -67,6 +67,21 @@ protected Iota(@NotNull IotaType<?> type, @NotNull Object payload) {
             HexEvalSounds.MISHAP);
     }
 
+    /**
+     * This method is called when this iota is executed inside parentheses (i.e. Hermes is run on a list containing Introspection, this, Retrospection).
+     * By default, the iota will be added to the in-progress parenthesized list instead of causing a mishap.
+     * This is specifically for parentheses-based escaping, Consideration escaping is handled in {@link CastingVM#executeInner}.
```
object-Object:
```suggestion
     * This method is called when this iota is executed inside parentheses (e.g. Hermes is run on a list containing [Introspection, this, Retrospection]).
     * By default, the iota will be added to the in-progress parenthesized list instead of causing a mishap.
     * This is specifically for parentheses-based escaping; Consideration escaping is handled in {@link CastingVM#executeInner} and cannot be overridden.
```

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/PatternIota.java
```diff
@@ -98,15 +122,42 @@ public boolean toleratesOther(Iota that) {
...
+                    throw new MishapInvalidPattern(this.getPattern());
+                }
             } else throw new IllegalStateException();
 
-            // do the actual calculation!!
-            var result = action.operate(
+            OperationResult result;
+            ResolvedPatternType resolutionType;
+            if (inParens) {
+                // handle parenthetized behavior
```
object-Object:
```suggestion
                // handle parenthesized behavior
```

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpCloseParen.kt
```diff
@@ -0,0 +1,45 @@
...
+            val newStack = image.stack.toMutableList()
+            newStack.add(ListIota(image.parenthesized.toList().map { it.iota }))
+            val image2 = image.copy(
+                stack = newStack,
+                parenCount = newParenCount,
+                parenthesized = listOf()
+            )
+            return OperationResult(image2, listOf(), continuation, HexEvalSounds.NORMAL_EXECUTE) to ResolvedPatternType.EVALUATED
+        } else if (newParenCount < 0) {
+            throw MishapNeedsParens()
```
object-Object: Is this case necessary? AFAIK `image.parenCount` can never be 0 here.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java
object-Object: Can we also update this comment to explicitly document that `vm.image.parenCount` is always greater than 0 when this is called?

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/Action.kt
```diff
@@ -41,6 +46,25 @@ interface Action {
         continuation: SpellContinuation
     ): OperationResult
 
+    /**
+     * The behavior of this action when inside parentheses. By default, this is just to add the pattern
+     * to the parenthesized list without updating the op count or performing any of its usual effects.
+     */
```
object-Object, line 56: Can we update this comment to explicitly document that `image.parenCount` is always greater than 0 when this is called?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpOpenParen.kt
```diff
@@ -0,0 +1,31 @@
...
+    override fun operateInParens(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation, thisIota: Iota): Pair<OperationResult, ResolvedPatternType> {
+        // we have escaped the parens onto the stack; we just also record our count.
+        val newParens = image.parenthesized.toMutableList()
+        newParens.add(CastingImage.ParenthesizedIota(thisIota, false))
+        val image2 = image.copy(
+            parenthesized = newParens,
+            parenCount = image.parenCount + 1
+        )
+        val resolutionType = if (image.parenCount == 0) ResolvedPatternType.EVALUATED else ResolvedPatternType.ESCAPED
+        return OperationResult(image2, listOf(), continuation, HexEvalSounds.NORMAL_EXECUTE) to resolutionType
```
object-Object: Can this be simplified to just `return OperationResult(...) to ResolvedPatternType.ESCAPED`, since `image.parenCount` is never 0 here?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpUndo.kt
```diff
@@ -0,0 +1,36 @@
...
+        val newParenCount = image.parenCount + if (last == null || last.escaped || last.iota !is PatternIota) 0 else when (last.iota.pattern.angles) {
+            HexActions.OPEN_PAREN.prototype.angles -> -1
+            HexActions.CLOSE_PAREN.prototype.angles -> 1
+            else -> 0
+        }
+        val image2 = image.copy(
+            parenthesized = newParens,
+            parenCount = newParenCount
+        )
+        // TODO: this should properly mishap if there was nothing to remove
```
object-Object: Do we have an issue for this? If not, can you create one?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpUndo.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    override fun operateInParens(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation, thisIota: Iota): Pair<OperationResult, ResolvedPatternType> {
+        val newParens = image.parenthesized.toMutableList()
+        val last = newParens.removeLastOrNull()
+        val newParenCount = image.parenCount + if (last == null || last.escaped || last.iota !is PatternIota) 0 else when (last.iota.pattern.angles) {
+            HexActions.OPEN_PAREN.prototype.angles -> -1
+            HexActions.CLOSE_PAREN.prototype.angles -> 1
+            else -> 0
+        }
```
object-Object: I know this was just copied from `CastingVM`, but can we refactor this? This is pretty unreadable. Probably something like the below would be easier to understand at a glance
```kt
var newParenCount = image.parenCount
if (last?.iota is PatternIota && !last.escaped) {
    when (last.iota.pattern.angles) {
        HexActions.OPEN_PAREN.angles -> newParenCount--
        HexActions.CLOSE_PAREN.angles -> newParenCount++
    }
}
```
(also, I'm fairly sure the `.prototype` is unnecessary here)

CHANGELOG.md
object-Object, line 1: Should we also document that Hasty Retrospection was renamed to Absent Introspection?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingImage.kt
```diff
@@ -72,6 +72,16 @@ data class CastingImage private constructor(
      */
     fun withResetEscape() = this.copy(parenCount = 0, parenthesized = listOf(), escapeNext = false)
 
+    /**
+     * Returns a copy of this with the provided iota added to the parenthetized list.
```
object-Object:
```suggestion
     * Returns a copy of this with the provided iota added to the parenthesized list.
```

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/Action.kt
```diff
@@ -41,6 +46,25 @@ interface Action {
...
+    @Throws(Mishap::class)
```
object-Object, line 57: Hmm, is it really a good idea to allow addons to throw mishaps while parenthesized? Do we handle this correctly? I don't think any base hex pattern currently does this.

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/Action.kt
```diff
@@ -41,6 +46,25 @@ interface Action {
...
+    fun operateInParens(
+        env: CastingEnvironment,
+        image: CastingImage,
+        continuation: SpellContinuation,
+        thisIota: Iota,
+    ): Pair<OperationResult, ResolvedPatternType> {
```
object-Object: Should we instead define a new `ParenthesizedOperationResult` that has all the same fields as `OperationResult` but also has `resolvedPatternType`? That would probably be a nicer API for addon devs. For `PatternIota`, we might also need to add an interface that they both implement, eg
```kt
interface IOperationResult {
  val newImage: CastingImage;
  // ...
}

data class OperationResult(
  override val newImage: CastingImage,
  // ...
) : IOperationResult

data class ParenthesizedOperationResult(
  override val newImage: CastingImage,
  // ...
  val resolutionType: ResolvedPatternType,
) : IOperationResult {
  constructor(operationResult: OperationResult, resolutionType: ResolvedPatternType) // ...convenience helper
}
```

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingImage.kt
```diff
@@ -72,6 +72,16 @@ data class CastingImage private constructor(
...
+     * [escaped] determines whether the iota should be considered escaped or not.
```
object-Object: Would consumers ever need to set `escaped` to `true`, or should we just remove it to avoid confusion? It seems like the *only* place this is used is for Evanition to know if the undone pattern was escaped by Consideration; and the relevant code in `CastingVM` doesn't use `withNewParenthesized` (I assume as an optimization).

While you're here, could you also document in `ParenthesizedIota` that that's what `escaped` is for?

object-Object (changes requested):
This is coming along! Overall this change looks pretty good, and it's nice to make the VM less hardcoded; however, there are some things in the API that could be cleaned up, as well as some potential refactors and documentation improvements.

Common/src/main/java/at/petrak/hexcasting/api/casting/castables/Action.kt
Robotgiggle: Throwing mishaps while parenthesized will be necessary to address the Evanition issue you mentioned above, along with a plan I have for a read-from-offhand-into-parens pattern. Also, I see no reason why allowing addons to do that would be problematic - imo it's better to natively allow interesting functionality like that rather than forcing people to make their own workarounds.

As for handling, any thrown mishap will be caught and handled by the try/catch in `Iota.lookupAndOperate`, exactly like a mishap thrown from the normal `Action.operate` method.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpUndo.kt
```diff
@@ -0,0 +1,39 @@
...
+import at.petrak.hexcasting.common.lib.hex.HexEvalSounds
+
+object OpUndo : Action {
+    override fun operate(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation): OperationResult {
+        throw MishapNeedsParens()
+    }
+
+    override fun operateInParens(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation, thisIota: Iota): ParenthesizedOperationResult {
+        val newParens = image.parenthesized.toMutableList()
+        val last = newParens.removeLastOrNull() ?: throw MishapNothingToUndo()
```
object-Object: IMO this shouldn't mishap; `Introspection, Evanition` should be a valid sequence of patterns to draw, resulting in a no-op

object-Object (approved):
LGTM!

PR #1049, opened 2026-05-17
Cache Slate Search results

c-Caelum:
Spell circles' slate search can take time to execute, and caching isn't expensive, so I figured I'd quickly implement caching. This was thrown together pretty quickly, and I'm open to any and all feedback you can give.


c-Caelum:
However, with the new optimisation of the slate search, this is rather negligible and is likely redundant.

I was originally going to make this a small addon with some other opinionated changes, but I decided against it because caching shouldn't have too many drawbacks.

Robotgiggle (commented):
Two potential issues:
- The cache gets invalidated if any BlockCircleComponent is placed or broken anywhere inside the AABB, so if you had a massive circle like a worldslate (which is exactly where you'd want to use the cache to avoid the enormous search time) the cache would be constantly getting invalidated by people building other circles inside it
- I don't know if this would actually have any performance impact, but running through the entire hashmap and checking whether the position is contained in any entries whenever any circle component is placed or broken anywhere seems like it could be kinda slow

c-Caelum:
Okay. I'll dwell over a way to fix these.

PR #1080, opened 2026-05-27
Change edified panel recipe to avoid conflict with vanilla chest

### LINKED ISSUE #964: edified panel recipe conflicts with vanilla chest
### Modloader

Fabric

### Minecraft version

1.20.1

### Hex Casting version

0.11.2-821ae79

### Modloader version

_No response_

### Modpack info

_No response_

### The latest.log file

_No response_

### Issue description

yup

Image

sometimes it's the other way around and you can't craft the panels, like so

Image

### Steps to reproduce

_No response_

### Other information

_No response_

Robotgiggle:
Fixes #964.

Also moves the `cannot_modify_cost` tag jsons from `src/generated/data` to `src/main/data` to stop them from being deleted by the datagen process. If there's some way to tell the datagen system to create an action tag with no entries, that would be a much better solution - please let me know if that's possible.


object-Object (approved):
LGTM; can you open an issue to look into the `cannot_modify_cost` issue more?

PR #1081, opened 2026-05-27
Update description of Inert Mindflay mishap

### LINKED ISSUE #1053: Inert Mindflay mishap description is outdated
### Minecraft version

1.20.1

### Hex Casting version

0.11.3

### Issue description

The description of the Inert Mindflay mishap states that it kills the entity no matter what, which is no longer true as of 1.20.1 hex. It should be updated to specify that only already-flayed things get killed, while other invalid recipes just deal 1 point of overcast damage.

Robotgiggle:
Fixes #1053


object-Object (approved):
LGTM

PR #1086, opened 2026-05-30
Fix thoth accumulator

### LINKED ISSUE #1085: Thoth accumulator no longer works
### Minecraft version

1.20.1

### Hex Casting version

0.11.3-pre-798

### Issue description

When using the latest build, Thoth's Gambit will always return an empty list no matter what was on the stack at the end of its iterations. This is probably caused by the changes from #1031 somehow.

Robotgiggle:
Fixes #1085. Now that the thoth accumulator is immutable and operations that previously updated it return a new instance instead, we need to actually return that new instance at the end of the thoth instead of returning the old unmodified one.

Also, the comment on line 59 was describing the wrong case, so I fixed it.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt
@@ -56,7 +56,7 @@ data class FrameForEach(
         level: ServerLevel,
         harness: CastingVM
     ): CastResult {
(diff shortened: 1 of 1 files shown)


s5bug (approved):
my bad OG https://github.com/Sirius902/NeoHexCasting/blob/d00e35d192f83450eabda34dc131893edfe29cc0/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameForEach.kt#L67

object-Object (approved):

PR #1089, opened 2026-06-04
Neoforge 1.21.1 stabilization

Mafsolin:
Summary
I put this together mainly for myself so I could get Hex Casting building again on NeoForge 1.21.1, but it may be useful for others too.

I used an AI coding agent for a large part of the migration/stabilization work, then verified the results locally and iterated on the build until it produced a working jar.

Verified
- `:Neoforge:compileJava`
- `:Neoforge:processResources`
- `:Neoforge:build`

Result
Build now produces a NeoForge 1.21.1 `0.12.0-devel` jar.


object-Object:
Our existing build already runs on NeoForge and is producing artifacts at https://ci.blamejared.com/job/petrakat/job/Hexcasting/job/1.21/. What changes does this PR actually make? It seems like just a lot of unnecessary permutations of Gradle.

object-Object:
After looking through the changes, it doesn't really seem like there's anything of substance here. As noted above, the existing code on our 1.21 branch already builds fine and produces public artifacts that you can directly download, so I'm not sure why this was necessary in the first place. We won't be merging this, and I'd advise others against trying to use it, especially since the scripts you've added are hardcoded to paths on your specific Windows machine.

Mafsolin:
You're right, I didn't notice that the existing 1.21 branch was already producing NeoForge artifacts in CI.

I started from local build/bootstrap issues on my side and overestimated how much of that was an upstream problem, so this PR ended up being much more about getting things reproducible in my environment than about fixing a missing upstream build target.

Sorry about the noise. I'm happy to close this PR, unless there are any small code-level pieces here that would be worth splitting out separately.

PR #1101, opened 2026-06-06
Fix entity iota comparison

### LINKED ISSUE #1100: Equality/Inequality Distillation will always treat entity iotas as different
### Modloader

Forge

### Minecraft version

1.21.1

### Hex Casting version

0.12.0-devel-pre-2

### Modloader version

Neoforge 21.1.230

### Modpack info

_No response_

### The latest.log file

https://mclo.gs/HUdJiAp

### Issue description

When comparing two entity iotas with either Equality or Inequality Distillation, the actions will always output as if the two iotas are different, even when the iotas are of the same entity

### Steps to reproduce

1. Draw Mind Reflection twice
2. Draw Equality Distillation
3. False is pushed instead of True

### Other information

_No response_

IridescentVoid:
Fixes #1100


Robotgiggle (approved):
LGTM

Issue #737, opened 2024-08-24
Improve documentation of iota embedding


object-Object:
Iota embedding should be more clearly documented. Currently the only explanation in the notebook is a brief mention under Hermes' Gambit, which is both an unintuitive location and misleading because it also works with Thoth, casting items, and so on.

Robotgiggle:
Fixed in #862

PR #1102, opened 2026-06-06
Add Contemplation pattern

### LINKED ISSUE #1092: Add Contemplation pattern
Consideration, but without escaped behaviour

IridescentVoid:
Adds Contemplation, a variant of Consideration that has no draw-time effect (and thus does not require double-escaping). Signature is 0.11.3 consideration but drawn in reverse.

Closes #1092


IridescentVoid:
~~For the book entry, I may also address #737 at the same time.~~

Edit: book rewrites consolidated into #1134

Robotgiggle (approved):
I added a changelog entry - otherwise, LGTM

PR #1103, opened 2026-06-06
Advanced parenthesis patterns (and they're Actions now)

Robotgiggle:
Adds Interjection, Meditation, and Recollection to allow more advanced pattern-list building.

- Interjection (used while introspecting) reads from the offhand and inserts the resulting iota into the in-progress list.
- Meditation pops an integer and opens a list to the specified parenthesis depth.
- Recollection closes a list from any parenthesis depth, pushing the depth and the list.

Meditation is treated as a normal pattern when inside existing parentheses because there's no way to determine how many new layers it should open. This is kind of a janky solution, but the only alternative would be to prevent using it inside parens altogether.

Recollection pushes the list on top of the depth for convenience, since you're much more likely to need the list as an argument for something (saving it to a focus, adding something with Integration Distillation, etc) than the number.

image

image


Robotgiggle:
FYI, I'm not entirely satisfied with the patterns for Meditation and Recollection - they're kind of awkward to fit into the grid when you have a bunch of other stuff. Recommendations would be appreciated, though make sure to avoid conflicting with stuff like Open/Close Transmit from Hexal.

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -1976,7 +1976,7 @@
...
-          "3": "This technique may be useful if I want a _Hex to be able to reference a specific iota, such as a complicated vector or an entity reference, without having to construct or obtain it each time.$(br2)The process of getting such an iota into a list of patterns in the first place may be somewhat involved. The simplest method would be to draw a placeholder pattern when assembling the list, and then make use of $(l:patterns/lists#hexcasting:replace)$(action)Surgeon's Exaltation/$ to replace it with my desired iota.",
+          "3": "This technique may be useful if I want a _Hex to be able to reference a specific iota, such as a complicated vector or an entity reference, without having to construct or obtain it each time.$(br2)To get such an iota into a list of patterns in the first place, I can either use $(l:patterns/advanced_escaping#hexcasting:read_into_parens)$(action)Interjection/$ while creating the list, or draw a placeholder pattern and then use $(l:patterns/lists#hexcasting:replace)$(action)Surgeon's Exaltation/$ to replace it once the list has been created.",
```
IridescentVoid, line 2026: It feels a bit weird to mention Interjection here. On a first read, the logical order would be to read 'Escaping Patterns' before 'Advanced Escaping'. We may want to consider shifting some of these extra commentary over to the other entry.

However, it's probably fine to merge this as-is, as I was planning to rewrite some of these when I implement #1107 anyway.

Robotgiggle: There are already a few instances of pages linking to other pages "further along" in the reading order (novice linking to bookkeeper, ravenmind linking to thoth, hermes linking to charon, etc) so I don't think this is necessarily an issue.

IridescentVoid: That sounds fair enough.
Edit: I would mark this as resolved, but github is 401ing despite showing me the button to do so.

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/FrameEvaluate.kt
```diff
@@ -17,7 +17,7 @@ import net.minecraft.server.level.ServerLevel
 /**
  * A list of patterns to be evaluated in sequence.
  * @property list the *remaining* list of patterns to be evaluated
- * @property isMetacasting only for sound effects, if this is being cast from a hermes / iris
+ * @property isMetacasting if this is being cast from a hermes / iris / thoth
```
object-Object: Nit: Can we make this a bit more generic? Presumably it should also be true for addon meta-eval patterns

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/PatternIota.java
```diff
@@ -167,10 +168,19 @@ public boolean executable() {
...
         } catch (Mishap mishap) {
+            // if a mishap happens mid-parens at the top level (ie when staffcasting), keep any in-progress parens
+            boolean wipeParens = false;
+            // if a mishap happens mid-parens while metacasting, wipe the parens - if we didn't do this, any open paren
+            // levels would remain open even though the metacast itself has mishapped and thus been aborted
+            if (continuation instanceof SpellContinuation.NotDone cnd && cnd.getFrame() instanceof FrameEvaluate frameEval) {
+                if (frameEval.isMetacasting()) {
+                    wipeParens = true;
+                }
+            }
```
object-Object: Can't we just replace all of this with a single inline expression? ie `boolean wipeParens = continuation instanceof ..... && frameEval.isMetacasting()`

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/PatternIota.java
```diff
@@ -167,10 +168,19 @@ public boolean executable() {
                 result.getSound());
 
         } catch (Mishap mishap) {
+            // if a mishap happens mid-parens at the top level (ie when staffcasting), keep any in-progress parens
+            boolean wipeParens = false;
+            // if a mishap happens mid-parens while metacasting, wipe the parens - if we didn't do this, any open paren
+            // levels would remain open even though the metacast itself has mishapped and thus been aborted
```
object-Object: Great comment!

Common/src/main/java/at/petrak/hexcasting/client/gui/GuiSpellcasting.kt
```diff
@@ -88,6 +85,21 @@ class GuiSpellcasting constructor(
         this.calculateIotaDisplays()
     }
 
+    fun canBeUndone(resPat: ResolvedPattern): Boolean {
```
object-Object: Nit: can we give `resPat` a less abbreviated name?

Common/src/main/java/at/petrak/hexcasting/client/gui/GuiSpellcasting.kt
```diff
@@ -88,6 +85,21 @@ class GuiSpellcasting constructor(
...
+        // standard escaped patterns can always be undone
+        if (resPat.type == ResolvedPatternType.ESCAPED) return true
+        // everything else cannot be undone, with three exceptions:
+        // - unescaped Introspection and Meditation can be undone if there's nothing else left to undo
+        // - Interjection can always be undone (undoing its inserted iota) despite using the EVALUATED coloring
+        if (resPat.type == ResolvedPatternType.EVALUATED) {
```
object-Object: Can we refactor this to use `when (resPat.type)`?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpOpenNParens.kt
```diff
@@ -0,0 +1,28 @@
...
+import at.petrak.hexcasting.api.casting.mishaps.MishapNotEnoughArgs
+import at.petrak.hexcasting.common.lib.hex.HexEvalSounds
+
+object OpOpenNParens : Action {
+    override fun operate(env: CastingEnvironment, image: CastingImage, continuation: SpellContinuation): OperationResult {
+        val newStack = image.stack.toMutableList()
+        val layers = when (val topIota = newStack.removeLastOrNull()) {
+            null -> throw MishapNotEnoughArgs(1, 0)
+            else -> listOf(topIota).getPositiveInt(0)
+        }
```
object-Object: Why not just use `getPositiveInt(newStack.lastIndex)` directly on `newStack`?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpReadIntoParens.kt
```diff
@@ -0,0 +1,47 @@
...
+        val (handStack) = env.getHeldItemToOperateOn {
+            val dataHolder = IXplatAbstractions.INSTANCE.findDataHolder(it)
+            dataHolder != null && (dataHolder.readIota(env.world) != null || dataHolder.emptyIota() != null)
+        }
+        // If there are no data holders that are readable, find a data holder that isn't readable
+        // so that the error message is more helpful.
+            ?: env.getHeldItemToOperateOn {
+                val dataHolder = IXplatAbstractions.INSTANCE.findDataHolder(it)
+                dataHolder != null
+        } ?: throw MishapBadOffhandItem.of(null, "iota.read")
```
object-Object: Nit: this indentation is pretty weird

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpUndo.kt
```diff
@@ -24,7 +24,7 @@ object OpUndo : Action {
         var newParenCount = image.parenCount
         if (last == null) {
             // if there was nothing in the parenthesized list, undo the initial open paren
```
object-Object: Can you update the comment to mention why we specifically need to set it to 0, not decrement it? Also, do we need more special handing for undoing Mediation/Recollection/Interjection? I almost wonder if we should update the new escaped-action API to add an undo handler method on actions so they don't need to be hardcoded here (and so addons can handle it too).

object-Object (changes requested):
This is a neat feature! The code is coming along, but I think the Evanition behaviour might need a bit more thought; also left some comments on smaller cleanup opportunities

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpReadIntoParens.kt
Robotgiggle: This is straight out of OpRead. That said, it is indeed quite weird, so I'll change it.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/escaping/OpUndo.kt
Robotgiggle: No additional handling is necessary as far as I'm aware:
- Opening with Meditation is handled by setting layers to 0 if the list was empty, and nested Meditation does nothing so it doesn't need special handling
- Recollection can't even be in an in-progress list (and thus be undo-able) unless it was considered, so no special handling needed there either
- Undoing an Interjection just removes whatever iota it inserted, and there's a line in GuiSpellcasting that allows the Interjection pattern to turn red like any other undone pattern

Robotgiggle: Actually, there was a bit of a weird interaction if you Interjected an open or close paren and then undid it, but I was able to fix that by just making Interjection set the `escaped` boolean on the ParenthesizedIota the same way Consideration does.

Common/src/main/resources/assets/hexcasting/patchouli_books/thehexbook/en_us/entries/patterns/advanced_escaping.json
```diff
@@ -0,0 +1,34 @@
...
+      "op_id": "hexcasting:open_n_parens",
+      "anchor": "hexcasting:open_n_parens",
+      "input": "num",
+      "text": "hexcasting.page.advanced_escaping.open_n_parens"
+    },
+    {
+      "type": "hexcasting:pattern",
+      "op_id": "hexcasting:close_all_parens",
+      "anchor": "hexcasting:close_all_parens",
+      "output": "num, [pattern]",
```
poolcritter: nit: doesn't iota embedding make this `num, [iota]`?

Robotgiggle: I think `list` is better than `[iota]` but yeah you're right about that

IridescentVoid (approved):
Looks good to merge, but has conflicts. We should wait for #1199 to land before trying to resolve the conflicts in the lang files.

object-Object (approved):

Issue #1111, opened 2026-06-07
Crash when mishapping with a list of iota


ChloeTax:
### Modloader

NeoForge

### Minecraft version

1.21.1

### Hex Casting version

1.12.0-devel-pre-2

### Modloader version

neoforge 21.1.233

### The latest.log file

https://mclo.gs/HsJAIsQ https://mclo.gs/MN5F43o

### Issue description

https://github.com/user-attachments/assets/8877c716-eb76-4b48-a2af-7aa7aee5107a
```
{
	{
	}
}
Numerical Reflection: 1
Division Distillation
```
but any other patterns inside the intro/retro work too

### Steps to reproduce

1. write a list of 2 iotas, like [minds, minds]
2. either
   - reveal the list to see mishap sparks
   - do a mishap including it for the pattern to grey out
   - do that mishap in a circle to crash the server

### Other information

This is with only hex and its dependencies, on a dedicated server

vgskye:
unknown if this happens on fabric(@c-Caelum couldn't reproduce on Fabric, apparently).

ChloeTax:
https://github.com/user-attachments/assets/77901e99-a3cb-434b-888b-436402115dff

https://github.com/user-attachments/assets/0ea69289-7155-4693-ad82-47ec255b8935
it seems to be fine on fabric

vgskye:
root cause: `HexConfig$ClientConfigAccess::alwaysShowListCommas` is unsound to have when the logical server may call `Iota::display`. either the config should be removed or all of serverside iota display needs to be refactored.

vgskye:
config was introduced in 6e04f5623ad4d01c61f4b97d17a1b43c1604659d as part of #844

Robotgiggle:
Fixed by #1113

PR #1112, opened 2026-06-07
Revert "Spell-list comma config toggle"

vgskye:
This reverts commit 6e04f5623ad4d01c61f4b97d17a1b43c1604659d.

see https://github.com/FallingColors/HexMod/issues/1111#issuecomment-4643661300 for rationale; the config is unsound as is and should not exist or be a common config since iota display is both-sided.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
@@ -4,7 +4,6 @@ import at.petrak.hexcasting.api.casting.SpellList;
 import at.petrak.hexcasting.api.utils.HexUtils;
 import at.petrak.hexcasting.api.utils.TreeList;
 import at.petrak.hexcasting.common.lib.hex.HexIotaTypes;
-import at.petrak.hexcasting.api.mod.HexConfig;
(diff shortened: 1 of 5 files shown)


Robotgiggle:
Wouldn't it also work to just skip the config check if we're not in the client? 

Making it a common config doesn't seem particularly useful since the point is for players to be able to use their preferred display style wherever possible.

vgskye:
> Wouldn't it also work to just skip the config check if we're not in the client?

no. that'll make it not crash but that'll make it silently misbehave since it will always act as if the setting is false(or true, depending on implementation) for iotas rendered on the server. moreover, a naive dedicated client check will make it also misbehave differently on LAN servers, following *host* behaviour instead of the expected client behaviour.

Robotgiggle:
That behavior seems preferable to completely removing it IMO - better to have customization in most places than none at all.

vgskye:
> That behavior seems preferable to completely removing it IMO - better to have customization in most places than none at all.

imo "sidedness of a particular instance of iota display" is not something that should be user-visible and thus anything that makes it visible is a no-go hack; I'll tag this as needs decision then

vgskye:
closing to reopen #1113 instead

PR #1114, opened 2026-06-07
Add iotatype comparison patterns

### LINKED ISSUE #569: Add type iotas or type predicates
cc @Alwinfy

IridescentVoid:
- [x] Add big equality and big inequality, testing for pattern type matching rather than value matching
- [x] Add corresponding notebook entries for the new patterns

Names are tentative, if there is a better name I am happy to switch it out.

Closes #569


Robotgiggle (approved):
LGTM

PR #1115, opened 2026-06-07
Fix Assessor's Purification for scrolls in item frames

### LINKED ISSUE #505: Both Assessor's patterns fail on scrolls
Writability is determined by checking if a null iota can be written to the item in question. Since you can't write null to a scroll, this always returns false.

Probably also applies to slates.

https://github.com/gamma-delta/HexMod/blob/60c55d86b1217da2288dc9b0ad8e8a177e63a5b8/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/rw/OpWritable.kt#L21

https://github.com/gamma-delta/HexMod/blob/60c55d86b1217da2288dc9b0ad8e8a177e63a5b8/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/rw/OpTheCoolerWritable.kt#L24

IridescentVoid:
Fixes #505

The behavior of Assessor's Purification:
- Item frame with focus - True (correct. before and after PR)
- Wall scroll entity - False (correct. before and after PR, though for different reasons)
- Item frame with scroll - False (incorrect. before) and True (correct. after PR)


object-Object (approved):
LGTM!

PR #1116, opened 2026-06-07
TreeList & Thoth's fix for 1.21

s5bug:
port of #1031 and #1086 to 1.21


vgskye (approved):
not reviewed: the dense 3kloc data structure implementation(I'm unqualified to review it)

assuming TreeList.java is good, everything else seems good

PR #1118, opened 2026-06-07
Hide arrow on argument-less pattern pages

### LINKED ISSUE #566: Hide arrow on pattern pages without arguments
image

Expected result: the arrow below the pattern should not be visible.

IridescentVoid:
Resolves #566

I don't quite like duplicating the `#text#` component with inverse guard clauses, but the alternative would be putting the y height in the PatternProcessor which feels worse.


Robotgiggle (approved):
LGTM

PR #1119, opened 2026-06-07
Add special velocity getter for item entities

### LINKED ISSUE #363: Pace Purification returns a vertical velocity when used on a nonmoving item
image
Version: 0.10.3 Fabric

IridescentVoid:
When on the ground and not moving horizontally, item entities only tick their movement every four ticks. However, velocity from gravity is still added every tick. Prior to this PR, that velocity would be reported by Pace Purification.

Fixes #363


Robotgiggle (approved):
LGTM

PR #1131, opened 2026-06-10
Fix HexPattern codec with graceful error handling

IridescentVoid:
Fixes #1128 

Root issue was the changes in #1120. This PR prevents upgrading from 1.21 pre-1120 to 1.21 post-1120 from bricking worlds, but pattern iota won't transfer across successfully.

I was expecting to need more changes, but it seems that great scrolls regenerate their data automatically (probably as a result of lazy creative scroll loading) and iota lists convert to garbage.

PR #1140, opened 2026-06-11
Fix HexPattern codec bugs caused by #1131

IridescentVoid:
#1131 accidentally broke pattern codec decoding and I missed that during testing, oops.
This PR actually fixes pattern decode now.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
@@ -119,7 +119,7 @@ data class HexPattern(val startDir: HexDir, val angles: MutableList<HexAngle> = 
         append("]")
     }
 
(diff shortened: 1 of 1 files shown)


Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
s5bug, line 132: Future: we were planning to move START_DIR back to the enum ordinal, right? Should probably make a note to test the intro/invalid/reveal stuff when that happens.

s5bug (approved):

Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
s5bug: I say this as having different type signatures for ANGLES and START_DIR would have prevented the original bug :P

Robotgiggle (approved):
LGTM

PR #1145, opened 2026-06-12
Linear scaling for impulse cost

Robotgiggle:
Even with the current system, you can easily get a linear cost for impulse by looping a bunch of power-1 impulses, and once you know about that trick there's no reason not to use it. The quadratic scaling basically just makes the spell more expensive to use for people who don't understand the implications of the (somewhat confusingly described) extra cost on additional casts.

I don't think it makes much sense to have an objectively better way to use a spell locked behind nothing but knowledge of a specific trick. Consequently, this PR removes the extra cost for multiple casts of Impulse on the same entity, and changes the cost formula itself to just be a simple `2*power`.

This is technically a significant buff, but anyone using high-level impulse for the last few years was using the linear cost trick anyway, so I don't think this change will actually have much effect on balance.


IridescentVoid (approved):
LGTM

PR #1153, opened 2026-06-12
fix mods.toml to list Curios dependency as Accessories is only used on Fabric

s5bug:
Ideally in the future the Xplat needed is removed but ATM10 at the very least has an explicit check + crash for if the Accessories Compatibility Layer is detected. So I think for now rather than moving 1.21 to Accessories on Forge the better solution is to just continue Xplatting it


s5bug:
Going to replace this with making Hex Casting actually use Accessories on Neoforge

PR #1159, opened 2026-06-14
Main to next

s5bug:
Includes PRs:
- #1115 
- #1081 
- #1080 

I'm not sure if I handled #1080 correctly. On this branch it becomes an addition rather than a rename.


Robotgiggle (approved):
LGTM! #1080 was indeed done correctly. The rename in the original PR was because the original version of the file would get deleted whenever datagen ran. At some point, somebody must have run datagen on the 1.21 branch and committed the result, removing the original file.

Robotgiggle:
(Also, sorry for taking so long to get to this)

PR #1161, opened 2026-06-14
Fix stale entity iotas read from holders

### LINKED ISSUE #1160: [1.21]Stale EntityIota read from iota holders can cause NPE and client disconnect
### Modloader

NeoForge

### Minecraft version

1.21.1

### Hex Casting version

1.21 branch build

### Modloader version

NeoforgeVersion:21.1.220

### Modpack info

_No response_

### The latest.log file

https://mclo.gs/AeIvi4A

### Issue description

Image

When an `EntityIota` is stored in an iota holder, such as a spellbook, thought-knot, or similar readable item, and the referenced entity no longer exists, reading the stored iota can still return the old entity iota for one evaluation step.

For example, if the entity has died, unloaded, or the referenced player is offline, using Scribe's Reflection / iota read on the holder still appears to read the old entity. The stack only updates to `Null` after the next pattern is drawn or evaluated.

This becomes more serious when the read happens during list evaluation. If the list reads an entity iota from the offhand item and then immediately uses it as an entity argument, the stale `EntityIota` can reach entity-consuming patterns before it is validated.

In a minimal development client, this may only produce an internal exception and make the spell fail. In a modpack environment, the same exception can additionally cause a `clientbound/minecraft:system_chat` encoding failure and disconnect the client.

Expected behavior: reading an iota holder should immediately normalize invalid entity iotas to `NullIota`. The client should not be disconnected by the mishap/error message.

Possibly related to #1097 and #1062, but this report is specifically about stale `EntityIota` values read from iota holders not being validated immediately, which can lead to an internal exception and a system chat packet encoding disconnect.

### Steps to reproduce

1. Store an entity iota in a spellbook, thought-knot, or another readable iota holder.
2. Make the referenced entity unavailable, for example by killing it, unloading it, or having the referenced player go offline.
3. Read the stored iota using Scribe's Reflection / iota read.
4. Observe that the old entity iota is still returned instead of `Null`.
5. Draw or evaluate another pattern.
6. Observe that the stack only updates to `Null` after the next validation step.
7. Optional: put the read operation inside a list evaluation and immediately use the result as an entity argument, such as with Blink.
8. Observe an internal exception. In some modpack environments this can also disconnect the client due to a `system_chat` packet encoding failure.

### Other information

_No response_

tozyx:
Fixes stale `EntityIota` values read from iota holders. Fixes #1160

When an `EntityIota` stored in a readable holder points to an entity that no longer exists, `OpRead` and `OpTheCoolerRead` returned the stale iota directly. The VM only validates the existing stack at the start of execution, so a newly-read stale entity iota could remain usable until the next validation pass.

This could cause entity-consuming patterns to receive an invalid `EntityIota`, leading to an internal `getEntity(...) must not be null` exception. In some environments, the resulting unknown mishap message could also fail to encode as a `system_chat` packet because the raw exception object was passed as a translatable component argument.

Changes:
- Validate iotas read from holders before returning them.
- Treat missing entities from `EntityIota` as `MishapEntityNotFound` instead of allowing a null entity to trigger an internal exception.
- Mark `EntityIota#getEntity` as nullable.
- Convert internal mishap exceptions to strings before using them in translated chat messages.

Tested with:

```text
./gradlew :Neoforge:build
./gradlew :Fabric:build
```
- Runtime-tested on NeoForge.
- Fabric was build-tested only.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/ActionUtils.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/ActionUtils.kt
@@ -4,6 +4,7 @@ package at.petrak.hexcasting.api.casting
 
 import at.petrak.hexcasting.api.casting.iota.*
 import at.petrak.hexcasting.api.casting.math.HexPattern
+import at.petrak.hexcasting.api.casting.mishaps.MishapEntityNotFound
 import at.petrak.hexcasting.api.casting.mishaps.MishapInvalidIota
 import at.petrak.hexcasting.api.casting.mishaps.MishapNotEnoughArgs
 import at.petrak.hexcasting.api.utils.asTranslatedComponent
@@ -37,7 +38,7 @@ fun List<Iota>.getDouble(idx: Int, argc: Int = 0): Double {
 fun List<Iota>.getEntity(level: ServerLevel, idx: Int, argc: Int = 0): Entity {
     val x = this.getOrElse(idx) { throw MishapNotEnoughArgs(idx + 1, this.size) }
     if (x is EntityIota) {
-        return x.getEntity(level)
+        return x.getEntity(level) ?: throw MishapEntityNotFound(x.entityId, x.entityName)
     } else {
         throw MishapInvalidIota.ofType(x, if (argc == 0) idx else argc - (idx + 1), "entity")
     }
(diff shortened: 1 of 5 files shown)


IridescentVoid:
This looks like a more fundamental issue with our EntityIota codec decoding, and we should fix it upstream if possible instead of adding checks in EntityIota consuming code.

PR #1164, opened 2026-06-14
Properly register media holder component on fabric

### LINKED ISSUE #1162: Casting items cannot find their own media reservoir
### Modloader

Fabric

### Minecraft version

1.21.1

### Hex Casting version

0.12.0-pre-16

### Issue description

On fabric 1.21 specifically, casting items give the not-enough-media mishap when used even if there's plenty of media available in their internal battery.

Robotgiggle:
Fixes #1162 by adding registration code for MediaHolderItem in the relevant part of FabricHexInitializer.

As a side note, I think much of the fabric-side item component stuff should be renamed now that it uses vanilla components rather than Cardinal Components, but that's out of scope for this PR (see #1165 for more).


IridescentVoid (approved):
Looks good to me

Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexInitializer.kt
s5bug, line 282: Do any of these need to be de-duplicated?

Robotgiggle: No, all of that is still necessary. The code I added is for handling items with their own classes that implement `MediaHolderItem`. All the items in that section don't have their own classes (they're just basic `Item` instances) so they can't extend `MediaHolderItem` and the media holder has to be registered manually.

s5bug (approved):

PR #1167, opened 2026-06-15
fix: readd deserialization-time validation for iota holders

### LINKED ISSUE #1160: [1.21]Stale EntityIota read from iota holders can cause NPE and client disconnect

IridescentVoid:
Fixes #1160 

Alternative implementation of #1161 that performs iota validation in the `readIota()` call where the iota component is deserialized instead of when OpRead/OpTheCoolerRead return. This should better ensure that the validation code gets called in the future as well.

I also opted to validate the stack to null newly disappeared entity iota after each frame, which should properly prevent anything from consuming invalid iota. This might have performance implications. Interestingly, 0.11.3 on 1.20.1 would have allowed operating on nonexistent/despawned entities as EntityIota used to hold Entity references directly.

Closes #1161

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
@@ -77,7 +77,11 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
                         sound = HexEvalSounds.MISHAP,
                     )
                 } else {
-                    result
+                    result.copy(
+                        newData = result.newData?.copy(
+                            stack = validateIotaList(result.newData.stack, world)
+                        )
+                    )
                 }
             }
 
--- a/Common/src/main/java/at/petrak/hexcasting/api/addldata/ADIotaHolder.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/addldata/ADIotaHolder.java
@@ -1,12 +1,13 @@
 package at.petrak.hexcasting.api.addldata;
 
 import at.petrak.hexcasting.api.casting.iota.Iota;
+import net.minecraft.server.level.ServerLevel;
 import org.jetbrains.annotations.Nullable;
 
 public interface ADIotaHolder {
 
     @Nullable
-    Iota readIota();
+    Iota readIota(ServerLevel world);
 
     @Nullable
     default Iota emptyIota() {
--- a/Common/src/main/java/at/petrak/hexcasting/api/addldata/ItemDelegatingEntityIotaHolder.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/addldata/ItemDelegatingEntityIotaHolder.java
@@ -3,6 +3,7 @@ package at.petrak.hexcasting.api.addldata;
 import at.petrak.hexcasting.api.casting.iota.Iota;
 import at.petrak.hexcasting.common.entities.EntityWallScroll;
 import at.petrak.hexcasting.xplat.IXplatAbstractions;
+import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.entity.decoration.ItemFrame;
 import net.minecraft.world.entity.item.ItemEntity;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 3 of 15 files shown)


IridescentVoid:
It looks like `unsealed` isn't used in any recipes, so I couldn't test the `test` function. Regardless, forge builds again.

Olfi01:
I think the update after every frame might lead to issues in multiplayer, as could any update without actually trying to read the iota.
For example when you've got a player iota stored somewhere and that player goes offline, the iota stored in the focus would immediately and irreversibly turn to null, or am I missing something?

IridescentVoid:
To be clear, "every frame" was referrering to every ContinuationFrame the CastingVM evaluates. Since it only happens as part of an active cast and only works on the stack, it shouldn't impact any inert storage mediums like iota-holding items.

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -77,7 +77,11 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
                         sound = HexEvalSounds.MISHAP,
                     )
                 } else {
-                    result
+                    result.copy(
+                        newData = result.newData?.copy(
+                            stack = validateIotaList(result.newData.stack, world)
```
IridescentVoid, line 82: We should also validate the ravenmind, since that is user-visible. Data internal to frames (such as the foreach accumulator) should be fine to leave as-is as it'd need to surface on the stack at some point for a hex to take advantage of it.

It may also be worth implementing caching for validation (since iotas are immutable) to reduce the performance impact of the eager checking.

Olfi01:
> To be clear, "every frame" was referrering to every ContinuationFrame the CastingVM evaluates. Since it only happens as part of an active cast and only works on the stack, it shouldn't impact any inert storage mediums like iota-holding items.

Oh, I see, that makes more sense :D

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
Olfi01: Can you maybe describe a bit more how you imagine the caching? From my understanding an entity could become unavailable at any time (logout, despawn, unload, death), and I don't really see a good way to invalidate a cache when that happens.

IridescentVoid: Oops, 3 am iris wasn't thinking straight. We can't cache things that hold volatile iotas. However, most iota types don't decay and always pass validation, so a list holding a large amount of vectors/numbers could still have validation skipped on subsequent passes which could help with performance still.

If the changes in this PR do not impact performance significantly, I think it would be fine to push caching to a later pull request (especially since this technically adds mutable internal state to Iotas, which I think is safe but would need to verify).

PR #1172, opened 2026-06-18
Make escaper patterns increment the op count

Robotgiggle:
When I turned the escaper patterns into `Action`s in #1047 I forgot that `Action` doesn't automatically handle incrementing the op count. This PR fixes that by manually incrementing it, just like in all the other patterns that directly extend `Action`.


IridescentVoid (commented):
Code changes seem to work as intended, but I'd like to discuss the balance implications before merging. As it currently stands, these changes significantly disadvantage using introspection and retrospection when nesting fragments of hexes.

Robotgiggle:
Part of the problem was that I made intro and retro increment the opcount even when inside a parenthesized list, which makes very little sense in retrospect. It not only balloons the op cost when deeply nesting but is also just plain unintuitive - normal patterns don't affect the opcount when parenthesized, after all.

I've removed that feature, so now intro and retro only consume ops when actually evaluated on the outer layer. Consideration still uses an op while parenthesized because the entire point of consideration is to have actual behavior when parenthesized instead of just going into the list.

IridescentVoid (approved):
Looks good to me.

PR #1178, opened 2026-06-23
Implement Fisherman's Gambit III

IridescentVoid:
Duplicates an iota in-place on the stack.

On a stack of 0, 1, 2, 1, casting Fisherman's Gambit III results in 0, 1, 1, 2.


IridescentVoid:
There were concerns about naming that should be addressed before this is merged. It's also worth questioning if this is a pattern worth implementing, or whether it'd be better as a special handler, etc.

PR #1183, opened 2026-06-30
Basic spells from Hexal

Robotgiggle:
Brings Smelt, Freeze, Falling Block, Particles, and Factorial Purification from Hexal into the base mod. This was discussed a while ago, during the 1.20 port, but it never really went anywhere. Now that we're porting again, and Hexal itself has been absorbed into the FallingColors organization, I think we should go ahead and do this.

There are two main motivations for this:
- These spells don't really fit with any of the main mechanics Hexal adds (wisps, links, and motes)
- There have been complaints that the base mod doesn't have enough useful spells, which this helps resolve


Common/src/main/java/at/petrak/hexcasting/api/utils/HexUtils.kt
```diff
@@ -64,6 +66,17 @@ fun vec2FromNBT(tag: LongArray): Vec2 = if (tag.size != 2) Vec2.ZERO else
         Double.fromBits(tag[1]).toFloat(),
     )
 
+fun FrozenPigment.nextColor(random: RandomSource): Int {
```
IridescentVoid: Why not put this on the FrozenPigment record directly? Additionally, a line in `FrozenPigment`'s javadoc reads: "Due to capabilities being really slow to query many times a tick on Forge, this returns a colorizer <i>supplier</i>. Get it once, and then query it a lot."
Is this line still correct on 1.21? If not, we should remove it. If it is, we probably should cache the colorProvider in `MsgSingleParticleS2C`/`MsgParticleLinesS2C`.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFallingBlock.kt
```diff
@@ -0,0 +1,128 @@
...
+import net.minecraft.world.phys.Vec3
+import kotlin.math.roundToLong
+
+// https://github.com/VazkiiMods/Botania/blob/1.21.1-porting/Xplat/src/main/java/vazkii/botania/common/item/lens/WeightLens.java
+object OpFallingBlock : SpellAction {
+    override val argc = 1
+
+    override fun execute(args: List<Iota>, env: CastingEnvironment): SpellAction.Result {
+        val pos = args.getVec3(0, argc)
+        env.assertVecInRange(pos)
```
IridescentVoid: Since this modifies the block, we should use `assertPosInRangeForEditing`.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFallingBlock.kt
```diff
@@ -0,0 +1,128 @@
...
+
+        val centered = Vec3.atCenterOf(BlockPos.containing(pos))
```
IridescentVoid: Why do we convert the `BlockPos` to a `Vec3`, pass it to the `Spell`, and then immediately convert it back to a `BlockPos`?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFallingBlock.kt
```diff
@@ -0,0 +1,128 @@
...
+            listOf(ParticleSpray.burst(centered, 1.0))
+        )
+    }
+
+    private data class Spell(val v: Vec3) : RenderedSpell {
+        override fun cast(env: CastingEnvironment) {
+            val pos = BlockPos.containing(v)
+
+            val blockstate = env.world.getBlockState(pos)
+            if (!env.canEditBlockAt(pos) || !IXplatAbstractions.INSTANCE.isBreakingAllowed(env.world, pos, blockstate, env.caster))
```
IridescentVoid: `canEditBlockAt` becomes redundant if we switch to `assertPosInRangeForEditing` in the action itself.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFallingBlock.kt
```diff
@@ -0,0 +1,128 @@
...
+                    10,
+                    0.45,
+                    0.45,
+                    0.45,
+                    5.0
+                )
+            }
+        }
+
+        fun canSilkTouch(level: ServerLevel, pos: BlockPos, state: BlockState, harvestTier: Tier, owner: Entity?): Boolean {
```
IridescentVoid: Is there no way for us to do this more neatly?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFreeze.kt
```diff
@@ -0,0 +1,64 @@
...
+import net.minecraft.world.level.material.Fluids
+import net.minecraft.world.phys.Vec3
+
+object OpFreeze : SpellAction {
+    override val argc = 1
+
+    override fun execute(args: List<Iota>, env: CastingEnvironment): SpellAction.Result {
+        val toFreeze = Vec3.atCenterOf(BlockPos.containing(args.getVec3(0, argc)))
+
+        env.assertVecInRange(toFreeze)
```
IridescentVoid: Switch to `assertPosInRangeForEditing`.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFreeze.kt
```diff
@@ -0,0 +1,64 @@
...
+                return
+            }
+            if (fluidState.type == Fluids.LAVA && blockState.block is LiquidBlock) {
+                env.world.setBlockAndUpdate(pos, Blocks.OBSIDIAN.defaultBlockState())
+                return
+            }
+            if (fluidState.type == Fluids.FLOWING_LAVA && blockState.block is LiquidBlock) {
+                env.world.setBlockAndUpdate(pos, Blocks.COBBLESTONE.defaultBlockState())
+                return
+            }
```
IridescentVoid: Is there a reason why these can't be data-driven?

IridescentVoid: Since flowing lava has an interaction, it feels weird not to give flowing water one too - frosted ice comes to mind though that melts into full source blocks.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpSmelt.kt
```diff
@@ -0,0 +1,104 @@
...
+        val result = optional.get().value.getResultItem(env.world.registryAccess()).copy()
+
+        if (result.isEmpty) return null
+
+        return result
+    }
+
+    private data class ItemSpell(val itemEntity: ItemEntity) : RenderedSpell {
+        override fun cast(env: CastingEnvironment) {
+            val result = smeltResult(itemEntity.item.item, env) ?: return // cursed .item.item to map from ItemEntity to ItemLike to ItemStack
```
IridescentVoid: Would it be better to do the recipe lookup in `execute` and mishap if the item is non-smeltable?

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpSmelt.kt
```diff
@@ -0,0 +1,104 @@
...
+
+            if (itemStack.item is BlockItem) {
+                env.world.setBlockAndUpdate(pos, (itemStack.item as BlockItem).block.defaultBlockState())
+
+                if (itemStack.count > 1) {
+                    itemStack.count -= 1
+                    env.world.addFreshEntity(ItemEntity(env.world, pos.x.toDouble(), pos.y.toDouble(), pos.z.toDouble(), itemStack.copy()))
+                }
+            } else {
+                env.world.destroyBlock(pos, false, env.caster)
```
IridescentVoid: `env.caster` is deprecated, should probably be `env.castingEntity`.

Common/src/main/java/at/petrak/hexcasting/common/casting/arithmetic/operator/OperatorUtils.kt
```diff
@@ -64,4 +68,27 @@ fun Iterator<IndexedValue<Iota>>.nextPositiveIntUnderInclusive(max: Int, argc: I
...
+/**
+ * Takes the standard factorial if `arg` is a positive integer, otherwise uses the gamma function.
+ */
+fun generalFactorial(arg: Double): Double {
+    val argInt = arg.roundToInt()
+    if (arg >= 0 && DoubleIota.tolerates(arg, argInt.toDouble())) {
+        // normal integer factorial
+        var result: Long = 1
+        for (factor in 2..argInt) {
+            result *= factor
```
IridescentVoid, line 82: This can overflow, and `21!` is enough to hit that case (for signed longs). This means we get different results compared to doing this with doubles.

IridescentVoid (commented):
Looked through the code and left a few comments. Haven't tested functionality in-game yet.

Common/src/main/java/at/petrak/hexcasting/common/casting/actions/spells/OpFallingBlock.kt
Robotgiggle: I was able to trim a little bit out, but not very much. I don't think there's any way around this general implementation concept, unless we do what Ars does and just use a blacklist tag instead - but then any mod that wants to be immune has to manually add their stuff to that tag, rather than having it automatically detected. 

The reason it's so involved in the first place is that we need to account for the various instances where a block will not drop itself, even with silk touch, unless you _also_ have the right tool and/or the right mining tier.

PR #1186, opened 2026-07-02
Fix mining tier check for Break Block, add hex_unbreakable tag

Robotgiggle:
Implements a much simpler method for checking whether a given block is breakable by a given tier, and moves the function itself out of the xplat system and into HexUtils. Also, adds an empty tag json for `hex_unbreakable`, which was referenced in the previous tier checking system but not actually implemented.


IridescentVoid (approved):
Looks good to me. Checks performed:
- Code looks logically sound.
- Compiles and runs fine.
- Break Block attempted on:
  - Bedrock - failure (silent) as expected
  - Sandstone - success as expected
  - With the following tag change:
    - `add(tag(HexTags.Blocks.HEX_UNBREAKABLE), HexBlocks.EDIFIED_PLANKS);` (placed in `HexBlockTagProvider.addTags` and ran datagen)
    - Edified Planks - failure (silent) as expected
  - With the following tag change:
    - `add(tag(BlockTags.INCORRECT_FOR_DIAMOND_TOOL), HexBlocks.EDIFIED_PLANKS);`
    - Edified Planks - failure (silent) as expected
    - Sandstone - success as expected

PR #1188, opened 2026-07-05
fix non-drawable patterns' stream codec

c-Caelum:
it used to use `fromAngles`, but it should be using `fromAnglesUnchecked`.


IridescentVoid (approved):
Looks fine to merge.

PR #1191, opened 2026-07-08
feat(api/iota): endofunctors

### LINKED ISSUE #1083: feat: add an iota walker / endofunctor method
### Describe the feature

add a function of signature `(Iota => Iota) => Iota` to iotas that recursively walks the tree, visiting and replacing all iotas visited. for most iotas (and likely for the default impl, if we provide one), this is just `def walk(f: Iota => Iota) = f(this)`. for lists, however, it would first apply f to all elements of the list, then pass a ListIota with the _new_ iotas to `f`. addons likewise could provide implementations for their container iota types.

### Additional context

hexal, for example, needs to walk the iota tree to handle replacing illegal iotas. currently, it relies on an explicit check for `ListIota`, then provides an `IMappableIota` extension point. however, this requires addon developers to be aware of hexal and requires them to add a dependency if they provide a container type. a default method on `Iota` would be much more discoverable, and wouldn't require adding an external dependency.

poolcritter:
adds `visit` and `visitChildren` to iotas letting you apply a UnaryOperator to an entire tree of them. this is like hexal's IMappableIota except I don't know if that does self pre or post (or if it does self at all). awa.

I haven't tested this in prod; hex doesn't have anything that would use this yet. @IridescentVoid I believe you have a patch that may want to use this?

Cc: @IridescentVoid
Change-Id: Ic464f4702ba9b0379c93e31723e2a5fe6a6a6964
Issue: closes #1083
Prereq: #1195

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
@@ -15,7 +15,9 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.function.UnaryOperator;
 
 import static java.lang.Math.max;
 
@@ -95,6 +97,18 @@ public class ListIota extends Iota {
     @Override
     public @Nullable Iterable<Iota> subIotas() {
         return this.getList();
+    }
+
+    @Override
+    protected Iota visitChildren(UnaryOperator<Iota> visitor) {
+        var out = new ArrayList<Iota>();
+        boolean conserve = true;
+        for (Iota orig : getList()) {
+            var replacement = orig.visit(visitor);
+            conserve &= replacement == orig;
+            out.add(replacement);
+        }
+        return conserve ? this : new ListIota(out);
     }
 
     @Override
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/Iota.java
@@ -18,6 +18,7 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
+import java.util.function.UnaryOperator;
 
 public abstract class Iota {
     @NotNull
@@ -100,6 +101,24 @@ public abstract class Iota {
      */
     public @Nullable Iterable<Iota> subIotas() {
         return null;
+    }
+
+    /**
+     * Applies the given operator to all iotas in this iota's tree (recursive subIotas), including this iota.
+     * @param visitor Applied to this iota then every element of the resulting iota's tree.
+     * @return The iota after all subiotas have been visited (and potentially replaced).
+     */
+    public Iota visit(UnaryOperator<Iota> visitor) {
+        return visitor.apply(this).visitChildren(visitor);
+    }
+
+    /**
+     * Applies the given operator to all iotas in this iota's tree, **excluding** this iota. This must be called on an
+     * iota just before it is returned from {@link Iota#visit}. You should try to preserve your iota's identity if
+     * the operator doesn't modify anything.
+     */
+    protected Iota visitChildren(UnaryOperator<Iota> visitor) {
+        return this;
     }
 
     /**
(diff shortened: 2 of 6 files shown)


IridescentVoid (changes requested):
github ate the comment I left on the file. see review below instead.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
```diff
@@ -97,6 +99,21 @@ public boolean toleratesOther(Iota that) {
...
+        int w = 0;
+        boolean k = true;
+        var p = getList();
+        while (p.getNonEmpty()) {
+            var r = visitor.apply(p.getCar());
+            k &= r == p.getCar();
+            o[w++] = r;
+            p = p.getCdr();
+        }
+        return k ? this : new ListIota(Arrays.asList(o));
```
IridescentVoid: This section is pretty hard to read. I would suggest the following snippet instead, which should be equivalent. Changes I've made:
- Gave variables longer names.
- Use explicit types instead of `var` for slightly better readability. (IMHO, SpellList being written directly is better.)
- Change the array to an ArrayList as `SpellList.size()` is O(n). This also eliminates a piece of state that needs tracking, the index, which helps with code understanding.
- Reordered some statements to follow what seemed to me as the most logical order.

```suggestion
        SpellList ptr = getList();
        ArrayList<Iota> updatedChildren = new ArrayList<>();
        boolean modified = false;

        while (ptr.getNonEmpty()) {
            var newChild = visitor.apply(ptr.getCar());
            updatedChildren.add(newChild);

            modified = modified || newChild != ptr.getCar();
            ptr = ptr.getCdr();
        }

        return !modified ? this : new ListIota(updatedChildren);
```

IridescentVoid (changes requested):
Code-Review: -1

Implementation of `visitChildren` should be made more readable. I would like to at least see full-length variable names.

I'd also like feedback from others on the API design since changing this later would be a breaking change.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
```diff
@@ -97,6 +99,18 @@ public boolean toleratesOther(Iota that) {
         return this.getList();
     }
 
+    @Override
+    protected Iota visitChildren(UnaryOperator<Iota> visitor) {
+        var out = new ArrayList();
```
IridescentVoid:
```suggestion
        var out = new ArrayList<Iota>();
```
~~Other than this, looks good to merge if there are no other comments on the api design.~~ Not good to merge, I'm pretty sure the logic is wrong.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
```diff
@@ -97,6 +99,18 @@ public boolean toleratesOther(Iota that) {
...
+        boolean conserve = true;
+        for (Iota orig : getList()) {
+            var replacement = visitor.apply(orig);
+            conserve &= replacement == orig;
```
penguinencounter, line 108: this is not kotlin so `==` means reference identity. are you sure that's right?

i suppose that worst-case you'll be rebuilding the iota unnecessarily

IridescentVoid: Comparison by reference is intended, since we don't want to be doing deep equality checks often. There's a doc comment to return the same instance when possible on `visitChildren`.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
```diff
@@ -97,6 +99,18 @@ public boolean toleratesOther(Iota that) {
         return this.getList();
     }
 
+    @Override
+    protected Iota visitChildren(UnaryOperator<Iota> visitor) {
+        var out = new ArrayList();
+        boolean conserve = true;
+        for (Iota orig : getList()) {
+            var replacement = visitor.apply(orig);
```
IridescentVoid:
```suggestion
            var replacement = orig.visit(visitor);
```

IridescentVoid (changes requested):
Code-Review: -1

Noticed a pretty major logic bug that I somehow missed during the initial review pass. We should probably test this more throughly before merging.

Request on using `ArrayList<Iota>` still applies.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/ListIota.java
poolcritter: > this is not kotlin so `==` means reference identity. are you sure that's right?

yes, we want similar behaviour to `mapConserve`

IridescentVoid (approved):
LGTM

IridescentVoid:
Also closes #1195, since the changes there are bundled here.

PR #1194, opened 2026-07-09
Add mishap-detection pattern

Robotgiggle:
Adds a system for mishap prevention by simulating patterns to see whether they'd mishap without having to actually cast them. Also removes the assessor/auditor patterns, since this makes them obsolete.

image


s5bug:
throwing `Shai's Consideration` as my hat in the ring in case a "cool" name is desired

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -150,12 +155,29 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
...
+            var result: CastResult
             if (this.image.parenCount > 0) {
                 // Handle parens escaping
-                return iota.executeInParens(this, world, continuation)
+                result = iota.executeInParens(this, world, continuation)
             } else {
                 // Handle normal execution behavior
-                return iota.execute(this, world, continuation)
+                result = iota.execute(this, world, continuation)
+            }
```
object-Object: Nit: this could be simplified using if expressions, eg
```
val result = if (this.image.parenCount > 0) {
    iota.executeInParens(this, world, continuation)
} else {
    iota.execute(this, world, continuation)
}
```

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -150,12 +155,29 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
...
+
+            if (this.image.simulateNext) {
+                // push a bool based on whether the cast would have succeeded; do not perform any side effects from the cast
+                val success = result.sideEffects.none { it is OperatorSideEffect.DoMishap }
```
object-Object: Would it be simpler to just check `result.resolutionType.success`, or is there some edge case where that wouldn't work?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingImage.kt
```diff
@@ -95,22 +97,24 @@ data class CastingImage(
                 Codec.INT.fieldOf("open_parens").forGetter { it.parenCount },
                 ParenthesizedIota.CODEC.listOf().fieldOf("parenthesized").forGetter { it.parenthesized },
                 Codec.BOOL.fieldOf("escape_next").forGetter { it.escapeNext },
+                Codec.BOOL.fieldOf("catch_mishap").forGetter { it.simulateNext },
```
object-Object: Should this be `simulate_next`?

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
```diff
@@ -150,12 +155,29 @@ class CastingVM(var image: CastingImage, val env: CastingEnvironment) {
...
+                newStack.add(BooleanIota(success && !tooBig))
+                val newImage = this.image.copy(
+                    stack = newStack,
+                    simulateNext = false
+                )
+                return CastResult(iota, continuation, newImage, listOf(), ResolvedPatternType.SIMULATED, HexEvalSounds.NORMAL_EXECUTE)
+            } else {
+                // return a standard CastResult that does include the side effects from the cast
+                return result
             }
```
object-Object: Nit: can we remove the `else`? it's logically equivalent and arguably a bit clearer

object-Object (approved):
LGTM with a bit of cleanup!

Common/src/main/java/at/petrak/hexcasting/api/casting/eval/vm/CastingVM.kt
Robotgiggle: Ah that's totally on me, forgot that was a thing

PR #1196, opened 2026-07-10
Add null-coalescing pattern

### LINKED ISSUE #326: Reimplement Old Conjunction/Disjunction [Suggestion]
The old conjunction and disjunction patterns were very useful for dealing with operations that may or may not produce Null, such as raycasts. Now that they're gone, the amount of patterns necessary to null-proof a raycast has fully doubled. Also, as pointed out by DaComputerNerd on the discord, C# and many other languages include operators that work exactly like the old conjunction/disjunction showing just how useful those operations are.

The reimplementation could be as two completely new patterns, or possibly as an alternate functionality for the existing ones based on whether your provide bools or anything else as input.

Robotgiggle:
Resolves #326 

image

Also slightly tweaks the wording for Augur's Exaltation to make it match the other logic operators (also reduces confusion imo).


poolcritter (approved):
no obvious issues from what I see

PR #1204, opened 2026-07-17
Use gnomecraft mirror for terraformersmc maven

object-Object:
This fixes the intermittent build failures caused by https://maven.terraformersmc.com/releases sometimes returning the following error:

```
Premature end of chunk coded message body: closing chunk expected
```

The new maven is a mirror run by a Terraformers team member.


Robotgiggle (approved):
LGTM

PR #1205, opened 2026-07-17
Use gnomecraft mirror for terraformersmc maven (1.21)

object-Object:
This fixes the intermittent build failures caused by https://maven.terraformersmc.com/releases sometimes returning the following error:

```
Premature end of chunk coded message body: closing chunk expected
```

The new maven is a mirror run by a Terraformers team member.

This implements the same fix as #1204, but for the `1.21` branch. This will cause a merge conflict when we next merge `main` into `1.21`, but it'll be trivial to fix, and it's currently more important to fix these failures to unblock other work.

### DIFF
--- a/Common/build.gradle
+++ b/Common/build.gradle
@@ -25,7 +25,8 @@ repositories {
 
     maven { url "https://maven.shedaniel.me/" }
     // EMI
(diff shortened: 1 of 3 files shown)


Robotgiggle (approved):
LGTM

PR #1208, opened 2026-07-24
fix: Extend truename protection to offline players

### LINKED ISSUE #1062: Enforce truename protections for offline players
It appears from inspection that the changes to `EntityIota` and `MishapOthersName` in the 1.21 port may have broken truename protections for offline players - i.e. it may be possible to save someone's truename if they're offline. We should first test to see if this is a real problem, then fix it if it is.

Original comment from #985: https://github.com/FallingColors/HexMod/pull/985#discussion_r2701760025

Olfi01:
Fixes #1062 

Even though I was only able to verify a possible exploit using another bug (list evaluation of stale iotas), it is still possible that an addon developer passes non-verified iotas to the `getTrueNameFromDatum` method.
To protect offline players as well, I added a boolean to the EntityIota like @slava110 suggested.

~This change will make all previously saved entity iotas turn into Garbage.~ Fixed

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
@@ -1,6 +1,7 @@
 package at.petrak.hexcasting.api.casting.iota;
 
 import at.petrak.hexcasting.common.lib.hex.HexIotaTypes;
+import com.mojang.serialization.Codec;
 import com.mojang.serialization.MapCodec;
 import com.mojang.serialization.codecs.RecordCodecBuilder;
 import com.samsthenerd.inline.api.InlineAPI;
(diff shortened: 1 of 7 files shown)


c-Caelum:
During deserialization, I suggest checking if it exists first. That way, previously saved iotas are not broken.

c-Caelum:
It fits nicely with the `validate` method as well, so you can default to true (that is, it is a player iota) and upon validation, update the iota to reflect that. At least, I think that will work.

Olfi01:
Thanks @c-Caelum :D

Olfi01:
I forgot to specify this in the comment as well: existing Entity Iotas wil nowl not turn into Garbage, I tested it :D
They are properly migrated now.

Robotgiggle (approved):
LGTM

PR #1210, opened 2026-07-24
chore: change ravenmind type to Iota on client side

### LINKED ISSUE #1059: Change `ExecutionClientView.ravenmind` from `CompoundTag?` to `Iota?`
In the 1.21 port, we changed `ExecutionClientView.stackDescs` from `List<CompoundTag>` to `List<Iota>`, but we left `ravenmind` as `CompoundTag?`. This is a weird API - we should make them both the same type. Since we can apparently now have `Iota` instances client-side, it's probably best to change `ravenmind` to `Iota?`.

https://github.com/FallingColors/HexMod/blob/bc65193f0c1262a8755926970e296b04d23137ee/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/ExecutionClientView.kt#L19-L20

https://github.com/FallingColors/HexMod/blob/bc65193f0c1262a8755926970e296b04d23137ee/Common/src/main/java/at/petrak/hexcasting/client/gui/GuiSpellcasting.kt#L45-L46

Original comment from #985: https://github.com/FallingColors/HexMod/pull/985#discussion_r2701736472

Olfi01:
Closes #1059 

Since Iotas can be used clientside after the 1.21 port, it makes sense to convert the Ravenmind to an Iota in all places for consistency.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/ExecutionClientView.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/eval/ExecutionClientView.kt
@@ -2,7 +2,6 @@ package at.petrak.hexcasting.api.casting.eval
 
 import at.petrak.hexcasting.api.casting.iota.Iota
 import at.petrak.hexcasting.api.casting.iota.IotaType
-import net.minecraft.nbt.CompoundTag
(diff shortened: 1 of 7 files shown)


Robotgiggle (approved):
LGTM

Issue #1072, opened 2026-05-27
Fix hexdoc build failures on 1.21


object-Object:
The hexdoc book is currently failing to build on 1.21 (see e.g. workflow runs for #985), and the `build_docs` workflow is not running on the `1.21` branch.

Olfi01:
From what I can see, at least one of the problems is the fact that there isn't yet a published version of hexdoc-minecraft for 1.21.1.
I don't know how much trouble it is to build and publish that, but without it it's probably hard to check for other problems.

PR #1211, opened 2026-07-25
Fix 1.21 CI

### LINKED ISSUE #1074: Fix failing pull request CI tests on 1.21
Most of the mc-runtime-test jobs in `ci.yml` are currently failing on 1.21 (see e.g. #985 workflow runs). We should fix these.

Olfi01:
Fixes #1074

There seemed to be multiple issues here:
- PAUCAL is relying on ArchitecturyAPI which was missing in the dependencies
- The PAUCAL version from SuperKnux' fork previously present in the dependencies caused an error on server start
- The unmapped *-shadow.jar files were copied to the mods folder together with the final mapped jar, causing the runs to non-deterministically fail depending on which file was loaded and which was ignored by the mod loader
- The mc-runtime-test action expects only the major java version, so it was failing with the full major.minor.patch format

I thought it was also worth reevaluating the mod caching of the testing workflow. This cache has thrown me for a loop multiple times while analyzing this issue, since the cached mods are all the mods present in the run/mods folder, not just the ones downloaded from the dependencies.txt. One possible way to avoid this would be to split the cache action call up into cache/restore and cache/save, saving the cache right after the dependency download instead of at the end of the job. I've included this solution into this PR, but feel free to let me know if this is not something you want.

Hexdoc still doesn't work, but this is tracked separately in #1072

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -94,9 +94,9 @@ jobs:
           ${{ matrix.dependencies }}
           END_OF_FILE
 
-      - name: Cache dependencies
+      - name: Restore cached dependencies
         id: cache
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
           path: run/mods
           key: ${{ hashFiles('dependencies.txt') }}
@@ -107,6 +107,13 @@ jobs:
           for url in ${{ matrix.dependencies }}; do
             wget --directory-prefix=run/mods/ "$url"
           done
+
+      - name: Cache dependencies
+        uses: actions/cache/save@v4
+        if: steps.cache.outputs.cache-hit != 'true'
+        with:
+          path: run/mods
+          key: ${{ hashFiles('dependencies.txt') }}
 
       - name: Download build artifacts
         uses: actions/download-artifact@v4
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -118,12 +118,18 @@ jobs:
           rm -f dist/*-javadoc.jar dist/*-sources.jar
           cp dist/hexcasting-${{ matrix.modloader }}-*.jar run/mods
 
+      - name: Shorten java version
+        shell: bash
+        id: shorten-java
+        run:
+          echo "shortened=${JAVA_VERSION%%.*}" >> $GITHUB_OUTPUT
+
       - name: Run MC test client
         if: matrix.env == 'client'
         timeout-minutes: 10
         uses: headlesshq/mc-runtime-test@4.5.1
         with:
-          java: ${{ env.JAVA_VERSION }}
+          java: ${{ steps.shorten-java.outputs.shortened }}
           mc: 1.21.1
           modloader: ${{ matrix.modloader }}
           regex: '.*${{ matrix.modloader }}.*'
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -115,15 +123,21 @@ jobs:
...
         run: |
-          rm -f dist/*-javadoc.jar dist/*-sources.jar
+          rm -f dist/*-javadoc.jar dist/*-sources.jar dist/*-shadow.jar
           cp dist/hexcasting-${{ matrix.modloader }}-*.jar run/mods
 
+      - name: Shorten java version
+        shell: bash
+        id: shorten-java
+        run:
+          echo "shortened=${JAVA_VERSION%%.*}" >> $GITHUB_OUTPUT
```
object-Object: Rather than this, could we just add a variable `JAVA_MAJOR_VERSION` at the top of the file right next to the `JAVA_VERSION` one?

.github/workflows/pr.yml
```diff
@@ -107,6 +108,13 @@ jobs:
             wget --directory-prefix=run/mods/ "$url"
           done
 
+      - name: Cache dependencies
```
object-Object, line 113: Nit: might be worth adding a comment about why this separate step is necessary

.github/workflows/pr.yml
```diff
@@ -107,6 +108,13 @@ jobs:
...
+        uses: actions/cache/save@v4
+        if: steps.cache.outputs.cache-hit != 'true'
+        with:
+          path: run/mods
+          key: ${{ hashFiles('dependencies.txt') }}
```
object-Object: Should we move the hash key into a variable so it's explicitly shared between the two cache steps?

object-Object (approved):
LGTM with a bit of cleanup for maintainability

.github/workflows/pr.yml
Olfi01: We could, sure, although my instinct is usually to avoid putting technically dependent configurations into two separate values

.github/workflows/pr.yml
Olfi01: Good idea, will do!

.github/workflows/pr.yml
Olfi01: Absolutely, I should have thought of that myself :D

.github/workflows/pr.yml
object-Object: I don't think it's realistically an issue - if they're directly next to each other, it's not reasonable to expect someone to change one and not notice the other one directly next to it. If you like, you can add a comment explicitly stating to keep them in sync

```yml
env:
  # Keep these two values in sync
  JAVA_MAJOR_VERSION: '21'
  JAVA_VERSION: '21.0.5'
```

object-Object: Anyway, the Jenkinsfile is only specifying the major version, so arguably GitHub Actions should just do the same:

https://github.com/FallingColors/HexMod/blob/46a16fc355cbc838d578cf0e79cee8a42956c6ee/Jenkinsfile#L6

.github/workflows/pr.yml
Olfi01: I used sha256sum instead of the GitHub Actions expression because I was unsure whether the latter is evaluated before the step itself.

Olfi01:
Wait, don't merge this, sha256sum seems to also return the filename again which is treated as a separate cache key, let me fix

Olfi01:
This still looks wrong. I'll have another look in a minute.

Olfi01:
Nevermind, it's fine, the sha256 + "dependencies.txt" cache entry just resulted in a partial cache hit which confused me. Should be all good to go now.

Issue #1179, opened 2026-06-23
Hex Casting 1.21.1 pre-16 conflicts with Freecam


704009905:
### Modloader

NeoForge

### Minecraft version

1.21.1

### Hex Casting version

hexcasting-neoforge-1.21.1-0.12.0-devel-pre-16.jar

### Modloader version

NeoForge 21.1.233

### Modpack info

A private modpack with lots of mods

### The latest.log file

https://mclo.gs/jiwd27C latest.log; https://mclo.gs/taD71Ed crash

### Issue description

When Hex Casting and Freecam are both installed, I can enter a world normally but when enable free camera it instantly crashes.

### Steps to reproduce

1. Install Hex Casting and Freecam on 1.21.1
2. Join a world (successfully)
3. Enter free camera mode (default F4)

Olfi01:
From what I could gather reproducing the bug myself, the issue arises because a RegisterBrewingRecipesEvent is fired when the fake player for Freecam is created, which throws an error in Hexcasting because it tries to access the potion registry, which is not accessible any more at this point.
I have not yet been able to determine which of the mods is doing something wrong. My guess is that Freecam is creating the fake player in a way that makes it rebuild all the recipes, which shouldn't happen after the world has started.

704009905:
> From what I could gather reproducing the bug myself, the issue arises because a RegisterBrewingRecipesEvent is fired when the fake player for Freecam is created, which throws an error in Hexcasting because it tries to access the potion registry, which is not accessible any more at this point. I have not yet been able to determine which of the mods is doing something wrong. My guess is that Freecam is creating the fake player in a way that makes it rebuild all the recipes, which shouldn't happen after the world has started.

Thanks for your information. I'm not familiar with Java, but I can add that I encountered several similiar crash (showing problems with registry minecraft:enchantments / minecraft:potions) when adding other mods with 1.21.1 Hex Casting. Most of those mods are fabric mods loaded by sinytra connector.
So maybe this problem is related to sinytra connector, or because I loaded Hex Casting on NeoForge.

Olfi01:
I was able to reproduce the problem without sinytra connector, so I don't think that is the issue. I believe it would be possible to build a workaround into Hex Casting, but I'm not sure whether that will be worth the effort, especially since #1065 might make the problem obsolete anyway.
[...]

PR #1212, opened 2026-07-26
chore: Consolidate classes to use IXplatRegister system

### LINKED ISSUE #1065: Consolidate all registry classes to use the new `IXplatRegister` system
We added a new `IXplatRegister` system and associated `IXplatAbstractions.INSTANCE.createRegistrar` in the 1.21 port; however, many existing registry classes (e.g. `HexBlocks`) still use the old manual `Map<ResourceLocation, T>` structure, and at least one new registry class (`HexStateIngredients`) doesn't even follow that convention. We should align *all* registry classes to use the new `IXplatRegister` system.

Original comment from #985: https://github.com/FallingColors/HexMod/pull/985#discussion_r2701874629

Olfi01:
Closes #1065 

As mentioned in that issue, a consistent way of registering things across the mod is desirable. With the IXplatRegister system, this can be achieved (as this PR does).

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/advancements/HexAdvancementTriggers.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/advancements/HexAdvancementTriggers.java
@@ -1,35 +1,26 @@
 package at.petrak.hexcasting.api.advancements;
 
+import at.petrak.hexcasting.xplat.IXplatAbstractions;
+import at.petrak.hexcasting.xplat.IXplatRegister;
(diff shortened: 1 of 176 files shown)


Olfi01:
No idea why the Fabric registries were lazily initialized when Forge registries are not, and Fabric is usually the eager one of the two anyway.

Olfi01:
This will also fix #1179, since the repeated event from Freecam will not cause an error any more.

Robotgiggle (approved):
LGTM

PR #1213, opened 2026-07-27
chore: cleanup unused action classes

### LINKED ISSUE #901: Remove unused action classes that are now implemented as arithmetic
When we added the new arithmetic system, many of the old implementations were left behind and are now unused. We should remove them to avoid confusion.

Technically this would be a breaking change, but I'd argue an addon depending on these classes at all would be a bug, since they aren't used for anything.

For example: [OpSub](https://github.com/FallingColors/HexMod/blob/6e065801b04072f6d02adfa44cfd70da582809ed/Common/src/main/java/at/petrak/hexcasting/common/casting/actions/math/OpSub.kt#L10)

Image

Olfi01:
Closes #901 

All the classes removed are unused in code and have been replaced by Arithmetic implementations.


Robotgiggle (approved):
LGTM

Issue #972, opened 2025-11-22
Remove PKPCPBP


object-Object:
Before the next release, we should figure out how to remove PKPCPBP from the Gradle build scripts and switch to https://modmuss50.github.io/mod-publish-plugin. PKPCPBP has some weird logic and requirements, and is generally just very fragile; and we can't easily update it.

PR #1215, opened 2026-07-28
chore: remove PKPCPBP and use mod-publish-plugin

DL909:
note that publish only triggers when last commit starts with "[release" (case-insensitive) 

also, due to the inconsistent of maven version and modrinth version, alongside modrinth strange and crude match mechanism, this version of build.gradle still needs manually specify more than a half of dependency's version.

### DIFF
--- a/Fabric/build.gradle
+++ b/Fabric/build.gradle
@@ -8,31 +8,54 @@ architectury {
     fabric()
 }
 
-// FIXME: update dependencies/versions before releasing 0.12
-pkSubproj {
-    pkPublish = true
-    curseforgeJar = remapJar.archiveFile
-    curseforgeDependencies([
-        "paucal",
-        "patchouli",
-        "fabric-language-kotlin",
-        "inline",
-        "cloth-config",
-        "cardinal-components-api",
-        "fabric-api",
-    ])
-    modrinthJar = remapJar.archiveFile
-    modrinthDependencies([
-        "paucal:dabyDTwJ", // 0.6.0-fabric
-        "patchouli:1.20.1-80-fabric",
-        "fabric-language-kotlin:1.13.7+kotlin.2.2.21",
-        "inline:1.20.1-1.2.2-fabric",
-        "cloth-config:11.1.106+fabric",
-        "cardinal-components-api:5.2.1",
-        "fabric-api:0.84.0+1.20.1",
-    ])
-}
+publishMods {
+    // for Fabric 1.21.11 and below
+    file = remapJar.archiveFile
+    modLoaders.add("fabric")
+    curseforge{
+        requires("paucal",
+                "patchouli",
+                "fabric-language-kotlin",
+                "inline",
+                "cloth-config",
+                "cardinal-components-api",
+                "fabric-api",)
+    }
+    modrinth{
+        requires {
+            slug = "paucal"
+            version = "6xwATRVY" // 0.7.0 fabric ver
+        }
+        requires {
+            slug = "patchouli"
+            version = "1.21-88-fabric"
+        }
 
+        requires {
+            slug = "fabric-language-kotlin"
+            version = fabricLanguageKotlinVersion
+        }
+
+        requires {
+            slug = "inline"
+            version = "1.21.1-1.2.2-fabric"
+        }
+
+        requires {
+            slug = "cloth-config"
+            version = clothConfigVersion + "+fabric"
+        }
+        requires {
+            slug = "cardinal-components-api"
+            version = cardinalComponentsVersion
+        }
+        requires {
+            slug = "fabric-api"
+            version = fabricVersion
+        }
+
+    }
+}
 
 loom {
     mixin.defaultRefmapName = "hexcasting.mixins.refmap.json"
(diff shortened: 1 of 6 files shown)


DL909:
#972

DL909:
I've tested this version on a temporary private modrinth project created for testing, and it works well. However, since I can't easily get a curseforge access token, I can't test the curseforge part.
also, since paucal 0.7.1 haven't release, I just use 0.7.0 as version. otherwise it will be refused by modrinth due to an invalid dependency.

build.gradle
```diff
@@ -14,38 +12,30 @@ plugins {
...
+def rootRecentCommit = ""
+def rootIsRelease = false
+try {
+    def proc = ["git", "log", "--pretty=tformat:%s", "-n", "1"].execute(null, rootProject.projectDir)
+    proc.waitFor()
+    rootRecentCommit = proc.text.trim()
+    rootIsRelease = rootRecentCommit ==~ /(?i)^\[release.*/
+} catch (Exception e) {
+    logger.warn("Could not determine release status from git: {}", e.toString())
+}
```
object-Object: Can we get rid of this and just have releases be triggered by manual Jenkins runs? I'm not a fan of having to push a commit to do releases. See an example here (though it'll need to be modified for Hex): https://github.com/object-Object/HexDebug/blob/8fc7e24e8313dc78433138100311fd81352716e3/Jenkinsfile#L11

build.gradle
```diff
@@ -14,38 +12,30 @@ plugins {
     // Also it looks like property lookups don't work this early
     id 'fabric-loom' version '1.9+' apply false
 
-    id("at.petra-k.pkpcpbp.PKPlugin") version "0.2.0-pre-104"
-    id("at.petra-k.pkpcpbp.PKSubprojPlugin") version "0.2.0-pre-104" apply false
-    id("at.petra-k.pkpcpbp.PKJson5Plugin") version "0.2.0-pre-104" apply false
+    id "me.modmuss50.mod-publish-plugin" version "2.1.1"
+    id "coffee.cypher.json-processor" version "0.1.0" apply false
```
object-Object: Actually, I'd like to still use PKJson5Plugin, just not the rest of PKPCPBP. Sorry, I should have made the issue more clear on that.

Fabric/build.gradle
```diff
@@ -8,31 +8,54 @@ architectury {
...
+    file = remapJar.archiveFile
+    modLoaders.add("fabric")
+    curseforge{
+        requires("paucal",
+                "patchouli",
+                "fabric-language-kotlin",
+                "inline",
+                "cloth-config",
+                "cardinal-components-api",
+                "fabric-api",)
```
object-Object: Nit: can we keep the previous format/indentation? Also applies elsewhere

Fabric/build.gradle
```diff
@@ -8,31 +8,54 @@ architectury {
...
+    modrinth{
+        requires {
+            slug = "paucal"
+            version = "6xwATRVY" // 0.7.0 fabric ver
+        }
+        requires {
+            slug = "patchouli"
+            version = "1.21-88-fabric"
+        }
+
```
object-Object, line 35: Nit: some unnecessary blank lines here

build.gradle
```diff
@@ -54,28 +44,55 @@ subprojects {
...
     }
 
-    pkJson5 {
-        autoProcessJson5 = true
-        autoProcessJson5Flattening = true
+//    pkSubproj {
+//        platform = project.platform
+//        artifactId = base.archivesName.get()
+//        versionDisplayName = "$platform-$minecraftVersion-$modVersion"
+//    }
```
object-Object: Can we get rid of this commented-out code?

object-Object (changes requested):
This is a good start, but I'd like to improve the publishing flow a bit more while we're here, and I'd still like to use PKPCPBP's JSON flattening. Also:
- Can this branch target 1.20 rather than 1.21? I'd like to get these improvements on 1.20 as well.
- We still need to support the Discord prerelease webhook that PKPCPBP (I think?) publishes to (the `#bleeding-edge-builds` channel in the VIP category). I believe mod-publish-plugin supports this, but if not, we may still need to use a bit of PKPCPBP code (this will likely require a parallel PR to PKPCPBP).

DL909:
> This is a good start, but I'd like to improve the publishing flow a bit more while we're here, and I'd still like to use PKPCPBP's JSON flattening. Also:
> 
>     * Can this branch target 1.20 rather than 1.21? I'd like to get these improvements on 1.20 as well.
> 
>     * We still need to support the Discord prerelease webhook that PKPCPBP (I think?) publishes to (the `#bleeding-edge-builds` channel in the VIP category). I believe mod-publish-plugin supports this, but if not, we may still need to use a bit of PKPCPBP code (this will likely require a parallel PR to PKPCPBP).

I will port it to 1.20 after finish those improvements.

and, the mod-publish-plugin do support this(discord), but I find nothing related to it when I rewrite build.gradle. I'll re-check it and hopefully migrate it to mod-publish-plugin, along with other improvements you mention.

In addition, I am not an experienced developer, and I am also not very proficient in English. Therefore, I hope you can understand any irregularities in the code and expressions. I am happy to learn from them and make improvements accordingly.

DL909:
@object-Object finished improvement. however, it seems that publish to maven is also handled by PKPCPBP, should we reproduce its logic or just ignore this lack? I'll port changes to 1.20 first right now.

PR #1202, opened 2026-07-16
Document head, tail, init, last in TreeList

s5bug:
Closes #1200 

This is how it looks in IJ:

{D55CE6B8-A345-48EE-AA61-3C0463DE6D05}

Not pretty, but not outright horrible...

PR #1216, opened 2026-07-28
Main to next

s5bug:
Includes PRs:
- #1115 
- #1081 
- #1080 
- #1171 
- #1147 
- #1169
- #1102
- #1204
- #1202
- #1213

Someone more familiar with the datagen than me needs to check why `scroll_paper.json` isn't being generated under the `Common/` or `Neoforge/` generated resources. `runAllDatagen` should probably completely delete the generated resources directory before running just so stuff like this can be caught at the right time.


Robotgiggle (approved):
Looks good, but I can't merge this myself since it needs to not be squashed. 

@object-Object

PR #1218, opened 2026-07-29
Add swords to Fabric tier-checking system to allow breaking cobwebs

Robotgiggle:
On Fabric, the check for "is Break Block able to break this" uses a list of tools and checks whether any of them are valid for the target block. However, cobwebs are broken using a sword, and that wasn't included in the list, which caused Break Block to be unable to break them. This PR fixes that.


object-Object:
Will this also affect things like long grass?

Robotgiggle:
Long grass and similar blocks already work fine, because they don't have the `requiresCorrectToolForDrops` boolean set to true like cobwebs do.

IridescentVoid (approved):
Looks good to me, and allows cobweb breaking as expected.

PR #1226, opened 2026-07-31
feat: automatically fix 1.20 items when switching to 1.21

### LINKED ISSUE #1214: NBT -> Component datafixer for updating existing 1.20 worlds to 1.21
### Describe the feature

Currently, if you update a world that was using 1.20 hex to 1.21, anything stored in custom NBT tags (casting items, iota holders, phials, etc) will vanish. The data is actually all still there, since it gets moved into the `minecraft:custom_data` component when you update the world, but it's not accessible ingame since 1.21 hex expects all that stuff to be in its own custom components.

We should implement some kind of datafixer to check for the 1.20-era custom NBT in the custom_data component and move it into the appropriate 1.21-era components, so that people can update existing worlds without losing all their stored hexes and iotas.

Olfi01:
Closes #1214 

Due to the switch from NBT data to components, items carried over from 1.20 to 1.21 don't retain stored data at the moment.
This PR implements an automatic fix for that, mixing into the process that vanilla Minecraft uses to do the same to its items.

As @IridescentVoid pointed out, a sample like this would probably also be a good addition to a 1.21 addon template.


Olfi01:
Waiting for #1220, otherwise this is ready

Robotgiggle (changes requested):
Leaving this here as a reminder: this system needs to be updated to account for the changes in #1270

PR #1230, opened 2026-08-01
feat: parse patterns from 1.20 entities & block entities correctly

### LINKED ISSUE #1225: Preserve 1.20 patterns on entities and block entities
### Describe the feature

At the moment, the codec for HexPattern isn't able to parse the numerical 1.20 way of serializing patterns.
Since one of the strengths of Codecs is their ability to support multiple versions of serialized data, it makes sense to implement an alternative parser that is able to parse the old structure.

### Additional context

_No response_

Olfi01:
Closes #1225

This PR allows the HexPattern codec to parse patterns in entities and block entities that were saved in 1.20 with the old numeric serialization.


Robotgiggle (changes requested):
The changes in #1270 mean this PR currently has a major merge conflict; you'll need to handle that (and make sure the legacy codec correctly converts to the new format) before it can get merged.

PR #1233, opened 2026-08-03
Fix issues with circle bounds

Robotgiggle:
At some point in the 1.21 port the `CircleExecutionState.bounds` value was changed from an `AABB` to a `BlockBox`. The `BlockBox` constructor does not need the one-block adjustment that the `AABB` constructor did in 1.20, so the presence of that adjustment was making the box one block too big.

Also, likely as part of the change to using `BlockBox`, the fold reflection patterns had been changed to return the actual corner positions of the bounds region, rather than the center of the corner blocks as in 1.20. This caused unexpected results when trying to use them to target blocks at the corners of the circle, since the game would have to determine which block to target given a position perfectly equidistant from 8 of them.

This PR fixes both of those problems, and also updates the guidebook descriptions for the fold reflections to clarify that they return the center positions of the corner blocks, rather than the actual corner positions.


Olfi01 (approved):
LGTM

s5bug (approved):

object-Object (approved):
Approving with write perms so GitHub will allow this to be merged; I haven't reviewed it myself

PR #1235, opened 2026-08-03
Pattern that pushes the amount of media available to the current cast

### LINKED ISSUE #858: Circle-only pattern that pushes the amount of media in the impetus
### Describe the feature

There's currently no (automatable) way to detect how much media is stored in an impetus. Adding comparator functionality wouldn't really help, since 99% of the time an impetus is nowhere even close to full. Instead, since there's already the precedent of circle-only spells like Waystone Reflection to get details about the current circle, why not add a new one that just pushes the amount of stored media to the stack as a number?

### Additional context

_No response_

IridescentVoid:
Closes #858

Pattern shape taken from hexpose with permission.


Robotgiggle (approved):
LGTM

PR #1236, opened 2026-08-04
Fix most kotlin warnings

### LINKED ISSUE #1044: Public copy() with private constructor deprecated in Kotlin
### Minecraft version

1.20.1

### Hex Casting version

PR #1043 

### Issue description

The new version of Kotlin used in the PR above gives a compilation warning whenever a data class has a public `copy()` method but a private constructor. In future versions of Kotlin this feature will be removed and the `copy()` method on data classes will always have the same visibility as the constructor. This affects us because `CastingImage` has a private constructor but makes very extensive use of a public `copy()` method.

The warnings can be suppressed using [this annotation](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-exposed-copy-visibility/), but that doesn't fix the underlying issue. A proper fix would be to either make the constructors of the relevant classes public, replace all usage of the auto-generated `copy()` method with our own custom method, or change the affected classes to not be data classes and then manually implement whatever auto-generated data class methods we need.

More info on the upcoming change and the deprecation process can be found on the documentation page for the annotation linked above, and also on the page for [this related annotation](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-consistent-copy-visibility/).

Robotgiggle:
Resolves #1044 by making the `CastingImage` constructor public. Also changes a bunch of `ctx: CastingEnvironment` parameters to be called `env` instead, both to fix some other warnings (about subclasses using different parameter names than their parents in overridden methods) and because it hasn't been called a CastingContext in years.

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/SpellAction.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/castables/SpellAction.kt
@@ -17,9 +17,9 @@ import net.minecraft.nbt.CompoundTag
 interface SpellAction : Action {
     val argc: Int
 
-    fun hasCastingSound(ctx: CastingEnvironment): Boolean = true
+    fun hasCastingSound(env: CastingEnvironment): Boolean = true
 
-    fun awardsCastingStat(ctx: CastingEnvironment): Boolean = true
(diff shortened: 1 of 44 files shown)


IridescentVoid (approved):
LGTM

PR #1237, opened 2026-08-05
fix: fabric spellbook scrolling

### LINKED ISSUE #1198: [1.21] Scrolling spellbooks broken on fabric
Can't scroll to change spellbook pages. Keybinds still work properly, so the issue is probably somewhere with the scroll hook.

Scrolling and keybinds both work properly on neoforge.

Olfi01:
Closes #1198 

The spellbook scrolling outside of the casting GUI was broken in Fabric due to a change in the Minecraft code, which resulted in the Mixin not being applied any more because it was unable to find a variable it required.
This PR fixes the Mixin and uses the more stable `@Local` capture from mixinextras instead of the base Spongepowered capture.


IridescentVoid (approved):
LGTM

PR #1239, opened 2026-08-07
make the jeweler's hammer able to be enchanted again

### LINKED ISSUE #1182: Jeweler's Hammer cannot be enchanted
### Minecraft version

1.21.1

### Hex Casting version

0.12.0-pre-16

### Issue description

In 1.21, the Jeweler's Hammer is no longer enchantable via either enchanting table or anvil. Attempting to apply books with an anvil does not work at all. Trying to use an enchanting table _does_ provide a list of potential enchantments (unlike most items which are truly non-enchantable) but none of the options offered are valid. On NeoForge, the options are displayed as disabled with a tooltip saying "limited enchantability". On Fabric, the options   appear normal but have no tooltip at all when hovered and cannot be clicked.

This behavior suggests that maybe the hammer is registered as enchantable somewhere but has not been made compatible with any existing enchantment.

NeoForge ench table behavior
Image

Fabric ench table behavior
Image

Anvil behavior
Image

s5bug:
Fixes #1182

### DIFF
--- a/Common/src/generated/resources/data/minecraft/tags/item/pickaxes.json
+++ b/Common/src/generated/resources/data/minecraft/tags/item/pickaxes.json
@@ -1,0 +1,5 @@
+{
+  "values": [
+    "hexcasting:jeweler_hammer"
+  ]
+}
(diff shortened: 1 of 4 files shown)


Olfi01 (approved):
LGTM

object-Object (approved):

PR #1240, opened 2026-08-07
fix missing pansexual pride colorizer recipe

### LINKED ISSUE #1130: [1.21] Port is missing recipe for hexcasting:pride_colorizer_pansexual
Recipe `hexcasting:pride_colorizer_pansexual` is missing from 1.21 port (intentionally commented as todo).
https://github.com/FallingColors/HexMod/blob/8c49a386e4137409636fed3883baf98ff50bb3c3/Common/src/main/java/at/petrak/hexcasting/datagen/recipe/HexplatRecipes.java#L213-L218

s5bug:
Fixes #1130

Adds a new tag `hexcasting:pans` with an optional entry for the `farmersdelight:skillet`


object-Object (approved):
These changes LGTM, but just curious about why the files are being generated both in Common and in Fabric/Neoforge. It seems like the generated files in Common are meant for Neoforge. Is that intentional?

s5bug:
No idea. I think that's just how runXplatDatagen or whatever works.

PR #1242, opened 2026-08-07
fix: properly track argument presence in the hex notebook

IridescentVoid:
Regression from #1118 

Prior to this fix, pages that had their header overwritten would not have their signatures visible on the book page. This affects Additive Distillation in the lists entry, and probably other entries.


Robotgiggle (approved):
LGTM

PR #1259, opened 2026-08-14
[1.21] Fixed & extended MishapInternalException

### LINKED ISSUE #1229: [1.21] internal exception kicks player from server due to message codec failure
https://github.com/FallingColors/HexMod/blob/6e14fae9510895a71b93d8970e91402700a5218a/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInternalException.kt#L19

when causing an internal exception, since the raw exception is passed to Component.translatable, server got something like
```log
[018��2026 13:20:38.836] [Netty Server IO #6/ERROR] [net.minecraft.network.Connection/]: Exception caught in connection
io.netty.handler.codec.EncoderException: Failed to encode packet 'clientbound/minecraft:system_chat'
	at TRANSFORMER/minecraft@1.21.1/net.minecraft.network.codec.IdDispatchCodec.encode(IdDispatchCodec.java:55) ~[client-1.21.1-20240808.144430-srg.jar%23234!/:?]
[......STACKTRACE OMITTED......]
Caused by: io.netty.handler.codec.EncoderException: Failed to encode: This value needs to be parsed as component translation{key='hexcasting.mishap.unknown', args=[dev.latvian.mods.rhino.EvaluatorException: Can't find method at.petrak.hexcasting.api.casting.eval.vm.CastingImage.copy(at.petrak.hexcasting.api.utils.TreeList$TreeList1,number,at.petrak.hexcasting.api.utils.TreeList$TreeList0,boolean,number,net.minecraft.nbt.CompoundTag). (startup_scripts:HEX/patterns/RegistryBase.js#75)]}
[......STACKTRACE OMITTED......]
```
, which results in an instant kick

YukkuriC:
fixes #1229
image

and adds detailed stacktrace (hover + click copy) for its debugging purpose
image

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInternalException.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/mishaps/MishapInternalException.kt
@@ -4,7 +4,13 @@ import at.petrak.hexcasting.api.casting.eval.CastingEnvironment
 import at.petrak.hexcasting.api.casting.iota.Iota
 import at.petrak.hexcasting.api.pigment.FrozenPigment
 import at.petrak.hexcasting.api.utils.TreeList
+import net.minecraft.network.chat.ClickEvent
+import net.minecraft.network.chat.Component
+import net.minecraft.network.chat.HoverEvent
+import net.minecraft.network.chat.Style
 import net.minecraft.world.item.DyeColor
+import java.io.PrintWriter
+import java.io.StringWriter
 
 class MishapInternalException(val exception: Exception) : Mishap() {
     override fun accentColor(ctx: CastingEnvironment, errorCtx: Context): FrozenPigment =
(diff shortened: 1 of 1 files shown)


object-Object (approved):
LGTM, but is there a particular reason this has to be for 1.21, or could it be merged into `main` instead so we can have this on 1.20 as well?

YukkuriC:
> LGTM, but is there a particular reason this has to be for 1.21, or could it be merged into `main` instead so we can have this on 1.20 as well?

yes it could, the primary reason for this PR is the kicking issue, and the stacktrace part is "we have to do manual toString anyway so why not" :3

object-Object:
Do we need to also merge this one, or is it ok to close this and merge #1268 into 1.21 in a main to next bundle?

YukkuriC:
> Do we need to also merge this one, or is it ok to close this and merge #1268 into 1.21 in a main to next bundle?

both ok for me :3

PR #1260, opened 2026-08-14
Caching pattern angles signature for high-frequency matches and else

YukkuriC:
tried my best to make its exposed interface stable, and relevant files change minimized ;w;

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
@@ -13,7 +13,32 @@ import net.minecraft.world.phys.Vec2
 /**
  * Sequence of angles to define a pattern traced.
  */
-data class HexPattern(val startDir: HexDir, val angles: MutableList<HexAngle> = arrayListOf()) {
+data class HexPattern(val startDir: HexDir, private val anglesMutable: MutableList<HexAngle> = arrayListOf()) {
+    // cached lazy
+    private lateinit var anglesLazy: Lazy<List<HexAngle>>
+    private lateinit var anglesSignatureLazy: Lazy<String>
+    fun resetCache() {
+        if (anglesLazy.isInitialized()) anglesLazy = lazy { anglesMutable.toList() }
+        if (anglesSignatureLazy.isInitialized()) anglesSignatureLazy = lazy(::anglesSignatureImp)
+    }
+
+    init {
+        resetCache()
+    }
+
+    // old interface
+    val angles: List<HexAngle>
+        get() = anglesLazy.value
+
+    fun anglesSignature() = anglesSignatureLazy.value
+
+    fun removeLast(): HexAngle {
+        if (anglesMutable.isEmpty()) throw NoSuchElementException()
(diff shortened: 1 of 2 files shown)


s5bug:
I think the answer for the better design here may be an immutable `HexPattern` and a mutable `HexPattern.Builder`. The angle signature for `HexPattern`s once they're built is almost always used, so it wouldn't have to be Lazy there either.

Additionally, if `HexPattern` is going to be a `data class`, its fields should be immutable anyway.

s5bug:
If we wanted to go super crazy, pattern building and checking could be a lot more optimized. I don't know Kotlin off the top of my head so I'll write the Java equivalents:

```java
public final class HexSignature {
  // array of integers that pack in angles
  // each int is split into bits:
  // ?? jjj iii hhh ggg fff eee ddd ccc bbb aaa
  // 000 → path ends before a multiple of 10 steps
  // 001 → forward
  // 010 → right
  // etc
  // this is so new int[n] initializes everything to "end of path" i.e. it doesn't have to be set manually
  // `10` should probably be put in a constant somewhere
  // keep in mind that arrays are mutable, so don't expose any setters
  private final int[] packedTurns;
  // memoized hashcode for fast lookup
  private final int memoizedHash;
  
  private HexShape(int[] packedTurns) {
    this.packedTurns = packedTurns;
    this.memoizedHash = Arrays.hashCode(packedTurns);
  }
  
  // override equals to use Arrays.equals
  // override hashCode to just return the memoizedHash
}

public final class HexPattern {
  private final HexSignature sig;
  private final HexAngle orientation;
  
  public static final class Builder {
    private final HexAngle startDirection;
    private final IntArrayList turnsBuilder;
  
    public Builder(HexAngle startDirection) { /* implementation omitted */ }
    
    HexPattern build() { /* implementation omitted */ }
  }
}
```

Calling `hashCode` on `HexSignature`s should be as fast as `String`s, and comparing them should be even faster (due to being smaller).

If we can assume that a vertex coordinate in a pattern will never exceed [-32768, 32767], then constructing a pattern while checking its angles can also be made a lot more efficient. Essentially you pack vertices into `int`s, and vertex pairs into `long`s. The trick is that you sort the `int`s before inserting them into your set, so you only have to store and check half the values:
```java
// a long[] actually may be more performant here!
// LongArraySet provides a convenient wrapper for something that can grow, but...
// new long[1 + (packedTurns.length * 10)] is guaranteed to be enough space to hold all the edges
LongOpenHashSet edgeSet = new LongOpenHashSet(1 + (packedTurns.length * 10));

// (q, r) coordinates for our current vertex
short fromQ = ...;
short fromR = ...;
// (q, r) coordinates for our next vertex
short toQ = ...;
short toR = ...;

int from = (fromQ << 16) | fromR;
int to = (toQ << 16) | toR;

// `if(to < from) /* swap */` is "technically" faster but definitely not worth the readability loss
int lesser = Math.min(from, to);
int greater = Math.max(from, to);
long edge = ((long) lesser << 32L) | (long) greater;

if(edgeSet.add(edge)) // idk
```

YukkuriC:
maybe someone else will continue this one ;w;

PR #1270, opened 2026-08-17
Efficient signatures for patterns

s5bug:
This needs an extra set of eyes and a bit of bug testing because I wrote all this while tired and hungry

Continuation of #1260, much thanks to @YukkuriC for the inspiration and motivation


s5bug:
I would like to mention that due to object padding, `HexSignature` actually has room on it to store a cached `Action`. _However_, I don't know if it's safe to do that: I don't know where serialization boundaries are / what would happen if someone rerolled per-world patterns while a Signature had an old Action cached.

If it would be safe to do that, please let me know so I can implement it :)

s5bug:
The other option (mentioned in Discord but not here) is to use that extra four bytes for `int size;` to not have to use `TURNS_PER_ARRAY_ELEMENT * packedTurns.length`. However, there's nothing that explicitly depends on the exact stroke count of a pattern to my knowledge.

Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexSignature.java
```diff
@@ -0,0 +1,276 @@
...
+                HexAngle inverseAngle = angle.inverse();
+                this.compass = this.compass.times(inverseAngle);
+
+                this.cursor = this.cursor.plus(this.compass.times(HexAngle.BACK));
+            }
+
+            return this;
+        }
+
+        public HexSignature build() {
```
s5bug, line 270: I don't have time to do it right now, but I need to remember to document properly that `build` can be called multiple times and `reset` can be used to not have to re-allocate.

Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexPattern.kt
```diff
@@ -1,207 +0,0 @@
...
-                Codec.STRING.fieldOf(TAG_START_DIR).forGetter(RawHexPattern::startDir)
-            ).apply(instance, ::RawHexPattern)
-        }.flatXmap({
-            val dir = HexDir.fromString(it.startDir)
-            try {
-                return@flatXmap DataResult.success(fromAnglesUnchecked(it.anglesSignature, dir))
-            } catch (exception: IllegalArgumentException) {
-                return@flatXmap DataResult.error { exception.message }
-            }
-        }, { DataResult.success(RawHexPattern(it.anglesSignature(), it.startDir.name)) })
```
s5bug, line 141: Old codec is incompatible with new codec

Common/src/main/java/at/petrak/hexcasting/api/casting/math/HexSignature.java
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "HexSignature[" + this.toAnglesString() + ']';
+    }
+
+    public static final Codec<HexSignature> CODEC = Codec.INT_STREAM
+            .xmap(IntStream::toArray, Arrays::stream)
+            .xmap(HexSignature::new, hs -> hs.packedTurns);
```
s5bug: Add a backup here that attempts to parse an anglesig produced by the old codec & mark with something (@Robotgiggle please advise) that identifies the old/backup codec as something to be removed before release / fixed as part of #1226

s5bug (commented):
Can't get them done right now but need to do these before merge

Robotgiggle (approved):
LGTM

PR #1273, opened 2026-08-20
added zero vector mention as per issue #1258

### LINKED ISSUE #1258: Zero vector is falsy but not documented as such
Hex notebook only lists 0, Null, False, and empty list as falsy, and says all other iota are truthy.

DetermiedMech1:
I didn't know whether i should also change the other languages, so i just changed en_us

fixes #1258


Robotgiggle (approved):
LGTM

PR #1275, opened 2026-08-23
fix: Stop adding knockback on overcasting damage

### LINKED ISSUE #1192: Healthcasting now deals random knockback
### Minecraft version

1.21.1

### Hex Casting version

0.12.0-pre-20

### Issue description

As of 1.21 hex, healthcasting damage now deals significant knockback in a seemingly random direction. This is both very annoying (can mess up archer/alidade casting, can make you pick up dropped media intended for use in a spell, etc) and does not match the pre-1.21 behavior.

### Additional info

A suggestion was made to add a unique damage sound for healthcasting damage in the process of fixing this.

Olfi01:
Fixes #1192 

A new damage type tag was introduced between 1.20.1 and 1.21.1 that determines whether a certain damage type deals knockback along with it. This PR adds that `#no_knockback` tag to the overcasting damge type.

As for the suggestion of a new sound for this damage type, I personally don't have a sound available or in mind, but I am open for input.

### DIFF
--- a/Common/build.gradle
+++ b/Common/build.gradle
@@ -13,11 +13,11 @@ repositories {
     // paucal, patchi, inline, JEI
     maven { url 'https://maven.blamejared.com' }
 
-    /*maven {
-        // location of the maven that hosts JEI files, temporarily disabled due to HTTP 523 errors
+    maven {
+        // location of the maven that hosts JEI files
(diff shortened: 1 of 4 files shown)


Robotgiggle (approved):
LGTM

PR #1276, opened 2026-08-23
chore: rename fromAngles and fromAnglesUnchecked

### LINKED ISSUE #1189: Rename HexPattern.fromAngles and fromAnglesUnchecked to make their intent more clear
In almost all cases outside of static pattern initializers, `fromAnglesUnchecked` is the better choice because it allows overlapping patterns; however, currently `fromAngles` (which throws for overlapping patterns) seems like the default choice since it's shorter.

Suggested names:
- `fromAngles`: `fromDrawableAngles`, `fromAnglesChecked`
- `fromAnglesUnchecked`: `fromAnyAngles`, `fromArbitraryAngles`

Olfi01:
Closes #1189 

I chose `fromDrawableAngles` and `fromAnyAngles` because to me (being relatively unfamiliar with the code) the option of "checked" wasn't clear on what is actually being checked and "arbitrary" is just a longer word that means essentially the same as "any".

### DIFF
--- a/Common/src/main/java/at/petrak/hexcasting/api/casting/arithmetic/Arithmetic.java
+++ b/Common/src/main/java/at/petrak/hexcasting/api/casting/arithmetic/Arithmetic.java
@@ -24,49 +24,49 @@ public interface Arithmetic {
 
 	// Below are some common Operator patterns that you can make use of in your Arithmetic:
 
-	HexPattern ADD = HexPattern.fromAngles("waaw", HexDir.NORTH_EAST);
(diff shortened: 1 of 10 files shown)


Olfi01:
The build seems to be failing due to some HTTP 523 errors from the JEI maven repo. Should probably be retryable once that is back up.

object-Object:
This may conflict with #1270

Olfi01:
> This may conflict with #1270 

Quite possibly, since it was an extremely simple change I'm happy to re-do it after that has been merged, if it remains necessary.

PR #1287, opened 2026-08-29
return paucal to maven coordinates & delete duplicate flatDir defns

s5bug:
Bruh, peep this one, like really peep this one. No more flatDir. No more libs. No more 19 dollar jars. No more trolls. You been blocked.


IridescentVoid (approved):
LGTM, builds and runs on both Fabric and Neoforge.

PR #1290, opened 2026-09-06
Only check for enlightenment when drawing outside parens

### LINKED ISSUE #1283: Drawing great spells within introspection while unenlightened fails
downstream issue ChloeTax/hexxy5#102
screenshot taken using the server modpack, running 0.12.0 pre-48

Image

Robotgiggle:
Fixes #1283


Common/src/main/java/at/petrak/hexcasting/api/casting/iota/PatternIota.java
s5bug, line 93: shouldn't this just be an early return if in parens?

s5bug (approved):

object-Object (approved):

PR #1299, opened 2026-09-13
Entity nulling rework

### LINKED ISSUE #1227: Entity iotas are not immutable after #1208
### Minecraft version

1.21.1

### Hex Casting version

0.12.0

### Issue description

All `Iota`s are supposed to be immutable (which is necessary for them to work properly as item component values) but the changes introduced in #1208 mean this is no longer true for `EntityIota`s. Specifically, the `isPlayer` field can be mutated by the `validate()` method when updating an entity iota that was created before this change, or when validating an iota for an offline player.

It's worth noting that the first scenario there will only ever happen once per iota, and in the second scenario the validation will fail and thus the `EntityIota` will be discarded in favor of a `NullIota`, but the fact remains that this _does_ still break the immutability contract for data components.

Robotgiggle:
Resolves #1227, removes the behavior of entity iotas becoming null when their referenced entity ceases to exist, and replaces it with an entity-doesn't-exist mishap thrown when you try to use an iota pointing to something nonexistent. To check whether the entity currently exists, you can use Augur's Purification on the iota.

Entity iotas nulling themselves is the only instance of iotas changing on their own without player input, so it was always a little bit unintuitive. However, naively removing that behavior would make it possible to save truenames of offline players. Previously, that problem was handled via an added `isPlayer` boolean on the iota, but this comes with its own problems (notably, making the iota mutable since setting the value on preexisting iotas needed to be done in `validate()` while the player in question was online). 

This PR implements a different solution that does not require any new data, codec shenanigans, or mutability. When the server starts up, it runs through its playerdata folder and stores all the UUIDs it finds into a set of "known players". When a new player joins, their UUID is also added to the set. Then, when checking playerness of an iota whose entity does not currently exist, it just compares the UUID to the set of known players. This does not account for players who have never joined the server before, but there's no way to get an entity iota for such a player without using commands, and with commands you can spawn in as many saved truenames as you want anyway.


s5bug (approved):

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
```diff
@@ -54,8 +56,10 @@ public Entity getEntity(ServerLevel level) {
         return entityName;
     }
 
-    public boolean isPlayer() {
-        return isPlayer;
+    public boolean isPlayer(MinecraftServer server) {
```
IridescentVoid: This method can be called a lot, so we should do internal caching of player-ness still (or check it only in the constructor, with a guard in case EntityIota is constructed on the client).

Example situations that result in a large number of disk read operations:
- Saving a list of entities from a zone distillation to an iota holder. One disk read per entity, for any type of entity.
  - Having a spell circle working with lists of entities.
- A Hexal cyclic wisp having a list of entities on its stack. Hexal is supposed to check each tick and adjust the upkeep cost based on whether player references are held. One disk read per entity per tick.

Common/src/main/java/at/petrak/hexcasting/api/casting/ActionUtils.kt
IridescentVoid: Stale references should give the proper mishap.

Also applies to the following methods in this file
- `getLivingEntityButNotArmorStand`
- `getMob`
- `getPlayer`

Common/src/main/resources/assets/hexcasting/lang/en_us.flatten.json5
```diff
@@ -1556,7 +1556,7 @@
         true_name: "I attempted to $(l:patterns/readwrite#hexcasting:write)$(action)save a reference/$ to another player to a permanent medium.$(br2)Causes black sparks, and robs me of my sight for approximately one minute.",
 
         "entity_not_found.title": "Entity Not Found",
-        entity_not_found: "I attempted to affect an entity that doesn't exist. Two cases for this; one is if the entity is simply dead. The other, and more peculiar to me, is if it is unable to be \"observed\" from afar.$(br2)Causes brown sparks, and causes me to feel sick for a moment, as if I had stood up too fast.",
+        entity_not_found: "I attempted to affect an entity that doesn't exist. This can happen if the entity is simply dead, or if it is unable to be \"observed\" from afar.$(br2)Causes brown sparks, and causes me to feel sick for a moment, as if I had stood up too fast.",
```
IridescentVoid: Nit: wording
```suggestion
        entity_not_found: "I attempted to affect an entity that doesn't exist. This can happen if the entity is dead or if it is unable to be \"observed\" from afar.$(br2)Causes brown sparks, and causes me to feel sick for a moment, as if I had stood up too fast.",
```

IridescentVoid (changes requested):
A few changes/blockers for merging:
- In discord: design discussion around intuitiveness and other points raised
- Likely performance issues in `EntityIota.isPlayer`. It would be best if we can run real benchmarks on whether this is a problem, including when the backing store is a hard disk rather than a SSD. Hex Casting should be playable even for players on older hardware.
- Incorrect mishaps raised by some `ActionUtils` helper methods (that were not updated by the pr, but probably should have been)
- Discussion: should cleric impeti allow binding to offline players, since references are always considered valid now? After all, binding impeti does not require ambit. (`BlockRedstoneImpetus.useItemOn`, around L73)

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
Robotgiggle: I've changed it to only read from the disk once on server startup, and to store the results of that into a hashset (which also gets updated when a new player joins). This way runtime checking is just as simple as seeing whether the iota's uuid is in the hashset, which should be quite a bit faster than reading the disk each time.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
```diff
@@ -93,31 +94,37 @@ private static Component getEntityNameWithInline(Entity entity) {
...
+            if (string.endsWith(".dat")) {
+                UUID playerId = UUID.fromString(string.substring(0, string.length() - 4));
+                playerUUIDs.add(playerId);
+            }
+        }
+    }
 
+    public static void addPlayerUUID(UUID newUUID) {
+        playerUUIDs.add(newUUID);
+    }
```
s5bug, line 120: I don't like these being public...

Fabric/src/main/java/at/petrak/hexcasting/fabric/FabricHexInitializer.kt
```diff
@@ -143,6 +145,8 @@ object FabricHexInitializer : ModInitializer {
         ServerTickEvents.END_WORLD_TICK.register(OpFlight::tickAllPlayers)
         ServerTickEvents.END_WORLD_TICK.register(OpAltiora::checkAllPlayers)
 
+        ServerPlayerEvents.JOIN.register { player -> EntityIota.addPlayerUUID(player.uuid) }
+
```
s5bug, line 149: ...but I don't see another way, considering how event handlers are registered :(

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
```diff
@@ -93,31 +94,37 @@ private static Component getEntityNameWithInline(Entity entity) {
...
 
+    // This finds anyone who has ever joined the server by reading the UUIDs from the playerdata folder
+    // We can't just use GameProfileCache because it only stores 1000 UUIDs and only keeps them for 30 days
+    public static void initPlayerUUIDs(MinecraftServer server) {
+        var playerDataDir = server.getWorldPath(LevelResource.PLAYER_DATA_DIR).toFile();
+        String[] uuidStrings = playerDataDir.list();
+        if (uuidStrings == null) return;
+        for (String string : uuidStrings) {
+            if (string.endsWith(".dat")) {
+                UUID playerId = UUID.fromString(string.substring(0, string.length() - 4));
```
s5bug: Minecraft pulls in Apache Commons Lang so this can be `StringUtils.removeEnd` for clarity

s5bug (approved):

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
```diff
@@ -102,7 +108,7 @@ public static void initPlayerUUIDs(MinecraftServer server) {
         if (uuidStrings == null) return;
         for (String string : uuidStrings) {
             if (string.endsWith(".dat")) {
-                UUID playerId = UUID.fromString(string.substring(0, string.length() - 4));
+                UUID playerId = UUID.fromString(StringUtils.removeEnd(string, ".dat"));
```
s5bug:
```java
try {
    UUID playerId = UUID.fromString(StringUtils.removeEnd(string, ".dat"));
    playerUUIDs.add(playerId);
} catch (NullPointerException | IllegalArgumentException ignored) {}
```

s5bug (changes requested):

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
Robotgiggle: The try-catch seems useful in case somebody puts something that isn't named a UUID into the playerdata folder, but I think the `endsWith(".dat")` check is still useful because most `<uuid>.dat` files will also come with a `<uuid>.dat_old`  file, and we don't need to throw and catch for all of those.

Common/src/main/java/at/petrak/hexcasting/api/casting/iota/EntityIota.java
```diff
@@ -74,7 +74,7 @@ public boolean toleratesOther(Iota that) {
 
     @Override
     public boolean isTruthy() {
-        return true;
+        throw new UnsupportedOperationException("To check truthiness, call getEntity() and see if it's non-null");
     }
```
s5bug, line 79: Why not make isTruthy take CastingEnv?

Robotgiggle: Feels weird to make a sweeping API change like that, especially since every other iota type doesn't need it.

s5bug: Making `isTruthy` not safe to call feels a lot worse though. Addon patterns that do selection will have to special-case it too.

Robotgiggle: ~~if I change the method signature addons will have to update their code anyway~~

That's a good point though, I'll change it

PR #1303, opened 2026-09-17
Minor tweak to OpLightning implementation

Robotgiggle:
As described on Discord [here](https://discord.com/channels/936370934292549712/940747085622419456/1549966643411681421), the current implementation of OpLightning doesn't allow the lightning bolt entity to be selected with things like zone distillations. It also doesn't currently attribute the bolt to the player who summoned it. This PR fixes both of those problems.


object-Object (approved):
These changes LGTM, but can we change this to target 1.20?

Robotgiggle:
1.20 version of the PR has been created (#1304).

